PR #20, opened 2015-05-03
unstable mode_t went to u32

cyisfor:
Since this crate attempts to emulate UNIX style permissions on #[cfg(windows)] but the unstable std::os::linux::raw::mode_t type is only going to be  available on #[cfg(linux)], it broke
when mode_t was changed from i32 to u32.

Easy enough to fix... u32 and i32 are treated the same when you write them to the tar file, anyway. Not as easy to detect if mode_t changes, if you never compile for Linux though.

I couldn't get the tests to run without #![enable(deprecated)] because of os::platform::fs::MetadataExt on line 310, but that probably should be fixed outside of the scope of this. With deprecated enabled, the tests passed fine for me (and I run #[cfg(linux)]).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -182,12 +182,12 @@ impl<R: Read> Archive<R> {
         return Ok(());
 
         #[cfg(unix)]
-        fn set_perms(perm: &mut fs::Permissions, mode: i32) {
+        fn set_perms(perm: &mut fs::Permissions, mode: u32) {
             use std::os::unix::prelude::*;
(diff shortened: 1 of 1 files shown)


cyisfor:
I tested this on the master git branch of rust and std:: as of today btw, so it might not be a problem until the next rust release. (When it *is* a problem, it breaks cargo!)

alexcrichton:
Folded in #21 (waiting for windows to pass), thanks!

PR #25, opened 2015-05-18
Fix directory traversal security issues with Archive::unpack().

cadencemarseille:
Similar to directory traversal security issues that have affected GNU tar in the
past, Archive::unpack() should not allow '..' to be used to overwrite arbitrary
files.

While I was in this area, I also made the following changes:
- Remove `#[cfg(feature = "nightly")]` and provide a fallback for unix and
  Rust stable that calls libc::chmod() instead.
- Fix the issue that set_perms() was not calling APIs that persisted the permissions
  changes. Permissions::set_readonly() does not modify the filesystem and presumably
  PermissionsExt::set_mode() does not, either.
- Ignore AlreadyExists errors from fs::create_dir_all().


examples/generate_malicious_tarball.rs
```diff
@@ -0,0 +1,40 @@
+extern crate tar;
```
alexcrichton: Perhaps this script could just be part of the test itself? It seems odd to have an example which generates a test case.

src/lib.rs
```diff
@@ -172,34 +169,93 @@ impl<R: Read> Archive<R> {
...
+            //   adjacent '/'s within the filename were consolidated into one
+            //   '/'.
+
             let mut dst = into.to_path_buf();
+            let mut contains_dot_dot = false;
+            let mut seen_non_empty_part = false;
             for part in bytes.split(|x| *x == b'/') {
+                if part == b".." {
+                    contains_dot_dot = true;
+                    break;
```
alexcrichton: I think this would be better expressed by removing the `contains_dot_dot` variable and instead labeling the outer loop with `'outer:` and then here saying `break 'outer`

alexcrichton: er, `continue 'outer`

src/lib.rs
```diff
@@ -172,34 +169,93 @@ impl<R: Read> Archive<R> {
...
             }
+            // If any part of the filename is '..', then skip over unpacking the
+            // file to prevent directory traversal security issues.  See, e.g.:
+            // CVE-2001-1267, CVE-2002-0399, CVE-2005-1918, CVE-2007-4131
+            if contains_dot_dot {
+                continue;
+            }
+            // Skip cases where only slashes were seen, because this is
+            // effectively an empty filename.
+            if !seen_non_empty_part {
```
alexcrichton: Right now the `std::path` module will not do anything whenever it sees `push("")`, so does this case need to be explicitly handled? Are the security-related implications associated with this?

src/lib.rs
```diff
@@ -172,34 +169,93 @@ impl<R: Read> Archive<R> {
...
+                libc::chmod(dst_cstring.as_ptr(), mode as libc::mode_t);
+                return Ok(())
+            }
         }
         #[cfg(windows)]
-        fn set_perms(perm: &mut fs::Permissions, mode: i32) {
+        fn set_perms(dst: &PathBuf, mode: i32) -> io::Result<()> {
+            let mut perm = try!(fs::metadata(dst)).permissions();
             perm.set_readonly(mode & 0o200 != 0o200);
+            try!(fs::set_permissions(dst, perm))
```
alexcrichton: Currently the build here is failing on windows because this requires an unstable feature unfortunately.

src/lib.rs
```diff
@@ -25,11 +25,9 @@ use std::io::prelude::*;
 use std::io::{self, Error, ErrorKind, SeekFrom};
 use std::iter::repeat;
 use std::mem;
+use std::path::{Path, PathBuf};
 use std::str;
 
-#[cfg(feature = "nightly")]
```
alexcrichton: I think for now it may not be worth trying to get `unpack` working on Nightly due to the Unix/Windows requirements

src/lib.rs
```diff
@@ -172,34 +169,93 @@ impl<R: Read> Archive<R> {
...
+            }
+            // Skip cases where only slashes were seen, because this is
+            // effectively an empty filename.
+            if !seen_non_empty_part {
+                continue;
+            }
+
             if is_directory {
-                try!(fs::create_dir_all(&dst));
+                if let Err(e) = fs::create_dir_all(&dst) {
```
alexcrichton: Perhaps this change could be separated out? I'd prefer to have a test case for this extraction ignoring the file already existing as well.

alexcrichton:
Awesome, thanks!

examples/generate_malicious_tarball.rs
cadencemarseille: Done

src/lib.rs
cadencemarseille: Done

src/lib.rs
cadencemarseille: I think that the only security issue from not skipping over essentially empty paths is that the permissions of the directory into which the tarball is extracted could be modified.

Note that in my latest patch, I also skip over '.'.

src/lib.rs
cadencemarseille: If possible, I would like to try to keep the changes to allow unpack to be used.  What I am trying to do is remove duplicated code from [libpcre-sys' build script](https://github.com/cadencemarseille/rust-pcre/blob/616d00a4075abc59ba571ee0a280068983891dd1/libpcre-sys/build.rs#L50).

src/lib.rs
cadencemarseille: The extracting_malicious_tarball() test depends on the second `if let Err(e) ...`.  What I have done in my new patch is add a test to check the first `if let Err(e) ...`.

src/lib.rs
cadencemarseille: In my latest patch, I added an `if cfg!(feature = "nightly")` guard to the not(unix) version of set_perms().

cadencemarseille:
I am not sure why the build is failing with Rust nightly:  
https://travis-ci.org/alexcrichton/tar-rs/jobs/63471092#L160

I am using rustc 1.2.0-nightly (0cc99f9cc 2015-05-17) (built 2015-05-17) and I do not see this error.

Do you know what the cause is?

alexcrichton:
Thanks again! Looked like the tests were passing on travis and just failing on Windows, so I pushed a few fixes. I also reverted the `if let Err(e) = ...` change for `create_dir_all` because I believe that the function `create_dir_all` already works if the directory already exists, so that error would just imply a race which shouldn't particularly be handled by this library I believe.

PR #28, opened 2015-07-28
Restore the mtime or the unpacked files

### LINKED ISSUE #27: tar-rs doesn't recover the mtime of the unpacked files
Current implementation of `tar-rs` does recover the file permission of the unpacked file but *not* for the `mtime`. `mtime` doesn't matter for most cases but sometime [it does][twitter].

target: https://github.com/alexcrichton/tar-rs/blob/68c7581006d829f79623846fcb303ef1a1cec178/src/lib.rs#L858
```rust
    fn unpack2(&mut self, dst: &Path) -> io::Result<()> {
        try!(io::copy(self, &mut try!(fs::File::create(dst))));

        // HERE will be the good position to restore mtime of the unpacked file.
        try!(set_perms(dst, try!(self.header().mode())));
        return Ok(());

        #[cfg(unix)]
        fn set_perms(dst: &Path, mode: u32) -> io::Result<()> {
            use std::os::unix::raw;
```

Unfortunately, the function [`set_file_times`][t] is unstable yet. So we may have to embed the function body of [`set_file_times`][t] into `tar-rs` just like you did with the [`cargo`][c].

##### Option A: Use `utimes` for unix, and `SetFileTime` for windows.
:+1: : I'll reuse the patch which you made before.
:-1: : Additional dependencies (`winapi` and `kernel32-sys`) will be added to `tar-rs` for windows.

###### Option B: Use [`utime`][u] for both unix and windows.
:+1: : No additional dependency.
:-1: : [`utime`][u] is [obsoleted][u2] by POSIX.1-2008.

I'm going to PR this. Do you agree with this plan? Which option do you prefer?

[twitter]: https://twitter.com/simnalamburt/status/625013067835813888
[t]: http://doc.rust-lang.org/stable/std/fs/fn.set_file_times.html
[c]: https://github.com/rust-lang/cargo/commit/23762e4586#diff-b99116b24344572ec62bb4702e213e50R119
[u]: http://doc.rust-lang.org/libc/libc/funcs/posix88/unistd/fn.utime.html
[u2]: http://man7.org/linux/man-pages/man2/utime.2.html#CONFORMING_TO

simnalamburt:
I did the utime before the chmod because some versions of utime are broken and trash the modes of the file. This behavior is equivalent with [GNU tar]. Also, I didn't just reused [the patch] you made before because of [a issue] with the Windows.

Closes #27 

[GNU tar]: https://github.com/williamholland/GNU-tar-mirror/blob/084a16e3d1/src/extract.c#L341
[the patch]: https://github.com/rust-lang/cargo/commit/23762e4586#diff-b99116b24344572ec62bb4702e213e50R119
[a issue]: https://github.com/simnalamburt/utime/commit/7a33507c5c5f


alexcrichton:
Hm actually on second thought, would you be ok folding functionality into the [`filetime`](https://github.com/alexcrichton/filetime) crate? That already has the equivalent of `get_file_times` and I suspect that adding `set_file_times` wouldn't be too hard. Some constructors would need to be added for `FileTime::from_seconds` perhaps, but it may be nice to consolidate file-time-related functionality into one crate.

Sorry I completely forgot about this crate earlier!

simnalamburt:
Oh no..

Actually I considered utime as a dead-simple file time library since designing time-related API was so hard for me. (e.g. file's local time vs utc, nanoseconds, should I use chrono? or other?, etc)

simnalamburt:
Oh no.. :cry: 

Actually I made `utime` as a dead-simple filetime library, since making well-designed time-related API was so painful. (should I make a new filetime struct of reuse `time` or `chrono`'s one? UTC filetime vs local filetime in Windows, cross platform issue with representing time because of Windows' weird `FILETIME` design, round-off error issue with converting `timespec` to `FILETIME` and vise versa, etc)

I took a look for the `filetime` and I had some points which I hesitate to agree with the API of it just as I did for the mine. (mainly with the platform specific `.seconds()`) And also personally, development of [`xz`][xz] is blocked because of the issue #27.

My suggestion:
1. Let's consider consolidating filetime related functionality into one crate as a one big another work.
2. Let's fix the issue #27 first for the others, and replace `utime` with the evolved `filetime` later.

I agree with your idea which there should be only one good filetime crate but it's getting harder for me to use more time into into this issue.

Sorry I didn't try to find existing filetime crate.

[xz]: http://github.com/simnalamburt/xz-rs

alexcrichton:
Ok no problem! I went ahead and added support to `filetime` and added a small commit on top of this to use that, thanks for the PR again!

simnalamburt:
:+1: for your kindness. You could close this PR and apply `filetime` directly but you didn't. I appreciate it.

PR #32, opened 2015-10-13
Document panicking on concurrent access

aidanhs:
Because append uses a refcell without synchronisation, I believe it can panic if concurrently accessed?


alexcrichton:
Hm could you elaborate more on what the failure scenarios you're thinking of here? Concurrent access is inherently forbidden because of `RefCell`, and synchronized access *should* always work out OK so long as the underlying object is `Seek`. I think that it's documented elsewhere about possible corruption or odd results if `Seek` isn't a bound.

aidanhs:
Yes, I misunderstood the behaviour of `RefCell`. Thanks for the explanation (again)!

PR #33, opened 2015-10-15
Add append directory method

aidanhs:
This (should) also allow passing directories through `append_path`.
There's a cost of one additional syscall when doing `append_path` with files :)

I've also
 - fixed an example
 - mentioned ustar
 - added a new test which attempts to do the same thing as the pre-prepared tar with directories

One question: I see that

> Note that this will not attempt to seek the archive to a valid position, so if the archive is in the middle of a read or some other similar operation then this may corrupt the archive.

> Also note that after all files have been written to an archive the finish function needs to be called to finish writing the archive.

is duplicated in four places now. Do you want it moving up to the top with each `append_` method just referrring people there?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -369,8 +369,16 @@ impl<W: Write> Archive<W> {
     }
 
     fn append_path2(&self, path: &Path) -> io::Result<()> {
-        let mut file = try!(fs::File::open(path));
-        self.append_file(path, &mut file)
+        let stat = try!(fs::metadata(path));
+        let header = try!(Header::from_path_and_metadata(path, &stat));
+        if stat.is_file() {
+            let mut file = try!(fs::File::open(path));
+            self.append(&header, &mut file)
+        } else if stat.is_dir() {
+            self.append(&header, &mut io::empty())
+        } else {
+            Err(Error::new(ErrorKind::Other, "path has unknown file type"))
+        }
     }
 
     /// Adds a file to this archive with the given path as the name of the file
@@ -397,7 +405,7 @@ impl<W: Write> Archive<W> {
     /// // Open the file at one location, but insert it into the archive with a
     /// // different name.
     /// let mut f = File::open("foo/bar/baz.txt").unwrap();
-    /// ar.append_path("bar/baz.txt").unwrap();
+    /// ar.append_file("bar/baz.txt", &mut f).unwrap();
     /// ```
     pub fn append_file<P: AsRef<Path>>(&self, path: P, file: &mut fs::File)
                                        -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


aidanhs:
Have now actually checked, it does allow passing directories through `append_path`.

src/lib.rs
```diff
@@ -369,8 +369,15 @@ impl<W: Write> Archive<W> {
     }
 
     fn append_path2(&self, path: &Path) -> io::Result<()> {
-        let mut file = try!(fs::File::open(path));
-        self.append_file(path, &mut file)
+        let stat = try!(fs::metadata(path));
+        if stat.is_file() {
+            let mut file = try!(fs::File::open(path));
+            self.append_file(path, &mut file)
```
alexcrichton: This may want to call `append` directly because the `Metadata` is already known (e.g. using `Header::set_metadata`)

src/lib.rs
```diff
@@ -415,6 +422,51 @@ impl<W: Write> Archive<W> {
...
+    /// use tar::Archive;
+    ///
+    /// let mut ar = Archive::new(Vec::new());
+    ///
+    /// // Open the file at one location, but insert it into the archive with a
+    /// // different name.
+    /// let s = fs::metadata(".").unwrap();
+    /// ar.append_dir("bardir", &s).unwrap();
+    /// ```
+    pub fn append_dir<P: AsRef<Path>>(&self, path: P, stat: &fs::Metadata)
```
alexcrichton: To mirror `append_path`, this may want to not take a `stat` perhaps?

src/lib.rs
```diff
@@ -415,6 +422,51 @@ impl<W: Write> Archive<W> {
...
+                                       -> io::Result<()> {
+        self.append_dir2(path.as_ref(), stat)
+    }
+
+    fn append_dir2(&self, path: &Path, stat: &fs::Metadata) -> io::Result<()> {
+        let mut header = Header::new();
+        // TODO: add trailing path::MAIN_SEPARATOR for compatibility when either
+        // the std::path allows it or OsStr permits character checks
+        // https://github.com/rust-lang/rust/issues/29008
```
alexcrichton: I think this should be possible to do by working with an `OsString` directly? This could call `.as_os_str()` and then do something platform-specific to test whether the last "piece" is a slash or not (and if so push a new one on).

aidanhs: Yes, I started down this route then got discouraged by how opaque OsStrs/OsStrings are to work with. In particular, I couldn't immediately see the correct way to get do the Windows-specific bit!
I plan to dive into OsStr(ing) a bit more to figure out how to use it properly, but it didn't seem important right now given my suspicion that all tar programs created in at least the last 15 years can handle ustar archives, so don't need the trailing slash.

src/lib.rs
aidanhs: I thought about this - what do you think the method should take? A path?
There's sadly no 'open directory' object.

src/lib.rs
aidanhs: It could, but then you've got the header building in two places? By calling `append_path` there's only one codepath for it.
I don't really have strong feelings - want me to move file header building into here?

This could also call the `append_{file,dir}2` functions instead of the public ones.

alexcrichton: Ah I was more going for the benefit of fewer syscalls rather than style here, `append_file` will call `metadata` again but we've already got the metadata here for the file. Currently the `foo2` functions are just the non-generic versions of `foo` to help codegen of downstream crates, so I wouldn't want to enhance the signatures too too much.

src/lib.rs
alexcrichton: Yeah I think a path is fine, it'll be a path to an existing directory which we learn the metadata from to insert into the archive.

src/lib.rs
alexcrichton: Oh if it's not required then this is totally fine :)

aidanhs:
I've made changes based on your comments.

Part of this was moving the special directory handling (size) into `set_metadata`. This overcomes the primary objection I had to passing append_directory a path, which was that there would be no good way to build a header/stat for a directory and reuse it for lots of different directory names. Because `set_metadata` now does the required cleverness, an API consumer can use that.

In addition, I created a helper function which allows (because of moving some cleverness into `set_metadata`) a significant reduction in code duplication across `append_{file,dir,path}`.

Thanks for your comments, I'm a lot happier with this now - the API surface is reasonable and there isn't any unnecessary duplication. Hopefully it looks a bit better to you?

I wasn't sure what the indent style should look like for `append_dir` given there now needs to be two trait bounds (I believe).

src/lib.rs
```diff
@@ -496,6 +548,11 @@ impl Header {
     /// file.
     pub fn set_metadata(&mut self, meta: &fs::Metadata) {
         self.fill_from(meta);
+        // Set size of directories to zero
+        let htype = self.link[0];
+        self.set_size(if htype == b'5' { 0 } else { meta.len() });
```
alexcrichton: Perhaps this could use `meta.is_dir()` instead of checking for `b'5'` (somewhat obscure)

alexcrichton:
Looking good to me! Just a minor comment about `stat.is_dir()` but otherwise I'm fine merging. Could you also squash down to a commit?

aidanhs:
Although it's not relevant any more, I have a question about this:

> Currently the foo2 functions are just the non-generic versions of foo to help codegen of downstream crates, so I wouldn't want to enhance the signatures too too much.

What did you mean by enhancing the signatures? As in, changing arguments? At that point in time, `append_path` would have actual `Path` objects so wouldn't need the generic function. Or are you saying that by calling the specific function it would interfere with Rust combining the generic+specific functions into one?

aidanhs:
Made the `meta.is_dir()` change, squashed to a single commit.

alexcrichton:
Ah yeah I just meant in terms of differing in signatures, nothing too major!

alexcrichton:
Thanks!

PR #34, opened 2015-10-17
Ensure invalid file sizes error immediately

aidanhs:
This fixes an infinite loop with file sizes which reach past the end of the file.


aidanhs:
(I'm also going to look at some changes to make the troublemaking infinite loop more robust in general)

aidanhs:
I found another way to trigger the infinite loop by just iterating over files, so I've added that to the test and made `skip` more defensive.

I think this means it's not actually required to check the result of io::copy any more, but it does give a clearer error message at the point that the error happens.

alexcrichton:
Thanks!

PR #35, opened 2015-10-17
Ensure individual blocks of 0s are not ignored

aidanhs:
http://pubs.opengroup.org/onlinepubs/009695299/utilities/pax.html gives no indication that individual blocks of zeroes should be allowed, nor can I see any evidence that the decision to silently ignore individual null blocks was deliberate in the history of this repo.

On a more practical note, tar on Ubuntu 14.04 doesn't like it either.

tar with two files, generated by the test with the null block line commented out:
```
$ tar tf thistar.tar
tar: Removing leading `/' from member names
/tmp/tar-rs.MWafrQyeTVvo/tmpfile1
/tmp/tar-rs.MWafrQyeTVvo/tmpfile2
$ echo $?
0
```

tar file generated by the test with the null block line included:
```
$ tar tf thistar.tar
tar: Removing leading `/' from member names
/tmp/tar-rs.c1NfuMTXreOI/tmpfile1
tar: A lone zero block at 2
$ echo $?
0
```
i.e. it stops reading at the null block. Not sure why the exit code is zero though!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -252,14 +252,19 @@ impl<R: Read> Archive<R> {
                  -> io::Result<Option<File<R>>> {
         // If we have 2 or more sections of 0s, then we're done!
         let mut chunk = [0; 512];
-        let mut cnt = 0;
         let mut me = self;
-        loop {
+        try!(read_all(&mut me, &mut chunk));
+        *offset += 512;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oops, very nice catch!

PR #37, opened 2015-10-22
Implement Clone, Eq and PartialEq for Header, expose as_bytes

aidanhs:
I was going to implement `Hash` as well, but I doubt there are many people who would use it.
Exposing as_bytes allows downstream crates to do it themselves if necessary.

If you prefer I can remove `Eq` and `PartialEq` and just let people implement those with `as_bytes` as well.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -523,6 +523,12 @@ impl<'a, R: Read> Iterator for FilesMut<'a, R> {
     }
 }
 
+impl Clone for Header {
+    fn clone(&self) -> Header {
+        Header { ..*self }
+    }
+}
+
 impl Header {
     /// Creates a new blank ustar header ready to be filled in
     pub fn new() -> Header {
@@ -551,7 +557,8 @@ impl Header {
     }
 
     /// Returns a view into this header as a byte array.
-    fn as_bytes(&self) -> &[u8; 512] {
+    pub fn as_bytes(&self) -> &[u8; 512] {
+        debug_assert_eq!(512, mem::size_of_val(self));
         unsafe { &*(self as *const _ as *const [u8; 512]) }
     }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -551,7 +568,7 @@ impl Header {
     }
 
     /// Returns a view into this header as a byte array.
-    fn as_bytes(&self) -> &[u8; 512] {
+    pub fn as_bytes(&self) -> &[u8; 512] {
         unsafe { &*(self as *const _ as *const [u8; 512]) }
```
alexcrichton: Could you also add a `debug_assert_eq!(512, mem::size_of_val(self))` here? Probably should have added that initially as well but if it's publicly exposed it's probably more important

src/lib.rs
```diff
@@ -523,6 +524,22 @@ impl<'a, R: Read> Iterator for FilesMut<'a, R> {
...
+    fn clone(&self) -> Header {
+        let mut header = Header::new();
+        unsafe { ptr::copy_nonoverlapping(self, &mut header as *mut Header, 1) };
+        header
+    }
+}
+
+impl PartialEq for Header {
+    fn eq(&self, other: &Header) -> bool {
+        self.as_bytes()[..] == other.as_bytes()[..]
```
alexcrichton: I think in theory this may want to discard fields like `_rest` and perhaps discard trailing bytes after the last 0 byte in fields like `name`, (e.g. how one might compare C strings). Perhaps this could do field-wise comparison via the accessors below?

src/lib.rs
```diff
@@ -523,6 +524,22 @@ impl<'a, R: Read> Iterator for FilesMut<'a, R> {
     }
 }
 
+impl Clone for Header {
+    fn clone(&self) -> Header {
+        let mut header = Header::new();
+        unsafe { ptr::copy_nonoverlapping(self, &mut header as *mut Header, 1) };
+        header
+    }
```
alexcrichton: I think this should be able to just be `Header { ..*self }`, right?

aidanhs: Oh neat! I need to re-read the rust reference.

src/lib.rs
aidanhs: Hmm, what would you suggest happens if a field access fails with error?
You raise an interesting point, because field comparison is not what I'd want but I can see others would.

Maybe `compare_{field,bytes}` methods which people can implement Eq on downstream if they want?

aidanhs: Actually, I'll remove Eq/PartialEq and come back to it when you can see my thoughts on a higher level api and extended headers etc.

alexcrichton: Oh right some of the numeric fields may fail with a parse error, although in those cases it may be possible to call `truncate` directly?

alexcrichton: I'd also be fine holding off on this for now

aidanhs:
Made changes. Appveyor seems broken?

alexcrichton:
Yeah hopefully the appveyor builds will come back tomorrow!

PR #38, opened 2015-11-16
Convert file api to entries api, use an enum for type flags

aidanhs:
The issue is that `Files` is actually an iterator over the entries of the archive, which (if created by GNU or posix archivers, i.e. anything in use today) may not *actually* be a file. I believe calling it `Files` is therefore not appropriate.

This is (should be) completely backwards compatible as type aliases have been added. The point of the first two commits being split ("Convert file api to entries api", "Convert tests to use new function names") is to prove that this is the case (and potentially help with bisection if something goes wrong).

The third commit isn't strictly related (it's just removing some of the hardcoded `b'5'` around the code), but it's part of the journey towards a library supporting GNU and posix extensions.

One note: I'm not particularly happy about having to have `EntryType::Unknown`, but it seems to be required for backwards compatibility (otherwise `fill_metadata` would need to be able to return an error if it was an unknown file type).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,50 +43,62 @@ macro_rules! try_iter{ ($me:expr, $e:expr) => (
 
 /// A top-level representation of an archive file.
 ///
-/// This archive can have a file added to it and it can be iterated over.
+/// This archive can have an entry added to it and it can be iterated over.
 pub struct Archive<R> {
     obj: RefCell<R>,
     pos: Cell<u64>,
 }
 
-/// An iterator over the files of an archive.
+/// Backwards compatible alias for `Entries`.
+#[doc(hidden)]
+pub type Files<'a, T> = Entries<'a, T>;
+
+/// An iterator over the entries of an archive.
 ///
 /// Requires that `R` implement `Seek`.
-pub struct Files<'a, R:'a> {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -41,52 +41,136 @@ macro_rules! try_iter{ ($me:expr, $e:expr) => (
...
+        EntryType::Fifo          => b'6',
+        EntryType::Continuous    => b'7',
+        EntryType::XHeader       => b'x',
+        EntryType::XGlobalHeader => b'g',
+        EntryType::GNULongName   => b'L',
+        EntryType::GNULongLink   => b'K',
+        EntryType::GNUSparse     => b'S',
+        EntryType::Unknown       => b' ',
+    }
+}
```
alexcrichton: Perhaps these could be inherent methods on `EntryType`? E.g.

```rust
impl EntryType {
    fn from_byte(byte: u8) -> Option<EntryType>;
    fn to_byte(&self) -> u8;
}
```

alexcrichton:
Sounds like a good idea to me, thanks! Could you also tag all the old methods with `#[doc(hidden)]`? That way they'll be squirreled away hopefully and can eventually be deprecated.

aidanhs:
Thanks for the tips, done.

PR #45, opened 2015-12-22
Add support for GNU long names

alexcrichton:


alexcrichton:
cc @aidanhs, this is what I was thinking so far.

Some thoughts of mine:

* Seems odd that `append` writes a GNU archive by default, but `entries` does not
* `GnuEntries::new` is pretty bad, will probably need to be something different
* I wonder if archives could have "modes of operation" where sometimes they operate in GNU mode and other times in raw mode? Perhaps raw mode could be the default?
* Otherwise implementation seems straigtforward enough.

I think I'd want to investigate a little more into how many fancy extensions there are to the tar format and try to scope out the design space here a bit as well.

mbrt:
Since `GnuEntries` seems to add some overhead (I've not measured it, but at least it's non-zero), I think it'd be a good thing to be able to switch from "raw" to "GNU" modes and vice-versa while iterating.

aidanhs:
@alexcrichton take a look at #46 for the kind of thing I was thinking of. I've shamelessly stolen a bunch of useful changes from this PR :) Please note that I've just thrown it together and it needs a number of edges sanding.

Brief comparison: rather than forcing you to identify that an archive is a GNU archive (that might be tricky as, it's valid to have posix extended headers as well as GNU extensions), the higher level `file_entries` method implemented there would abstract away that detail and magically handle GNU extension for you. POSIX extended headers can be implemented with a hashmap on both FileEntriesMutFields (i.e. the iterator) (which would be used to accumulate global posix headers, type 'g') and FileEntryFields (i.e. FileEntry internals) (which would be used to hold the combination of global posix headers and item posix headers, type 'x').

> I wonder if archives could have "modes of operation" where sometimes they operate in GNU mode and other times in raw mode? Perhaps raw mode could be the default?

As mentioned above, because it's possible to mix and match different extensions I'd want a high level reader API to just magically handle it all for me - this is how the golang tar package works. It's an open question how one would decide what format you want to *write* in.

src/gnu.rs
```diff
@@ -0,0 +1,109 @@
...
+macro_rules! try_iter {
+    ($e:expr) => (match $e {
+        Ok(e) => e,
+        Err(e) => return Some(Err(e)),
+    })
+}
+
+/// dox
+pub struct GnuEntries<'a, R: 'a> {
+    inner: Box<Iterator<Item=io::Result<Entry<'a, R>>> + 'a>,
```
mbrt: Why are you using a boxed `Iterator` here? Can't you use a fully generic struct to avoid virtual dispatch?

src/gnu.rs
```diff
@@ -0,0 +1,109 @@
...
+        };
+
+        if !entry.header().entry_type().is_gnu_longname() {
+            return Some(Ok(GnuEntry { inner: entry, name: None }))
+        }
+
+        // Don't allow too too crazy allocation sizes up front
+        let cap = cmp::min(entry.header().size().unwrap_or(0), 128 * 1024);
+        let mut filename = Vec::with_capacity(cap as usize);
+        try_iter!(entry.read_to_end(&mut filename));
```
mbrt: This could read more than the `Vec` capacity, so a crazy allocation is still possible. Isn't it? I think this should be replaced by `entry.read`.

alexcrichton: Yeah it's definitely possible that this could go off the rails, which is a concern of @aidanhs's I believe for handling things such as malicious archives.

src/gnu.rs
alexcrichton: I personally prefer to keep this as non-generic as possible to decrease monomorphization costs

alexcrichton:
I've been looking into supporting other forms of archives (like pax) and perhaps some other GNU extensions, so I'll want to tweak this I think a bit. Regardless, though, I think we all agree on there should be two modes of operation:

* A low-level interface providing raw, unparsed access to each of the entries.
* A higher-level interface which performs decoding such as what is added to this PR.

I personally think the higher-level interface should continue to be the default as it's likely what new users want, but I'm curious what you guys think?

mbrt:
@alexcrichton I agree with you. I think most of the users don't even know what those formats are. I also think that basic operations should be easier than advanced. And dealing with different formats is definitely an advanced operation. For this reason the higher-level interface should be the default.

alexcrichton:
As an update, I'm still working on this (slowly) and will likely incorporate much of @aidanhs's [ideas](https://github.com/alexcrichton/tar-rs/pull/46). I hope to send a PR this week or this weekend to get some feedback on this. I'm going to take this opportunity to make some other breaking changes and release 0.4 as well.

alexcrichton:
Ok, I've summed up what I was thinking in https://github.com/alexcrichton/tar-rs/pull/50, which subsumes this.

PR #46, opened 2015-12-23
Add support for GNU long names (alternative)

aidanhs:


alexcrichton:
Ok, I've summed up what I was thinking in https://github.com/alexcrichton/tar-rs/pull/50, which I believe covers this, pax extensions, and continues to strike a reasonable balance between being low-level yet usable. Please let me know what you think, however!

Issue #48, opened 2015-12-30
Handle pax_global_header


dbrgn:
With the released version (0.3.2) of tar-rs, when unpacking a Github `.tar.gz`, the directory contains a file called `pax_global_header`. See https://github.com/backdrop-ops/contrib/issues/55.

So I tried the current master, because I saw that you worked on headers. Now I get the following error:

```
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: Error {
repr: Custom(Custom { kind: Other, error: TarError {
desc: "failed to unpacked `foo/pax_global_header`",
io: Error { repr: Custom(Custom { kind: Other,
error: StringError("unknown file type 0x67") }) } } }) }', ../src/libcore/result.rs:738
```

I'm not sure how to deal with that, is some code missing to handle this header?

dbrgn:
Found it in the tar specs: https://www.gnu.org/software/tar/manual/html_node/Standard.html (`XGLTYPE`). I'll try to come up with a PR.

alexcrichton:
A very good point! I recently looked a bit more into the various flavors of archives that can exist, and I'm thinking of tweaking the APIs here and there to support all the flavors I could find (including pax here).

I hope to send a PR soon!

SoloKarma:
https://github.com/alexcrichton/tar-rs/issues/48

Issue #39, opened 2015-11-23
Expose the entry type in Header


mbrt:
The entry type is now parsed from `Header`, but the information is kept private. The raw information is available trough the public field `link`, so the information is already there. I think it's useful to:
* promote `EntryType` to be public;
* add a `entry_type` getter in `Header`.

alexcrichton:
This seems pretty plausible to me, yeah, although I'd prefer to tweak the API if this were the case:

* The returned value should be an opaque struct instead of an `enum`
* Variants should be moved to `is_foo` accessors.
* There'd be a method to extract the raw underlying byte.

Other than that though I'm game!

aidanhs:
> The returned value should be an opaque struct instead of an enum

Out of curiosity, why have you gone this route? This means match can't check for exhaustiveness any more.

aidanhs:
In particular, users of this library now won't be informed automatically if I add support for posix extended headers and gnu extensions - they'll need to read the library documentation, rather than the compiler informing them on library upgrade.
(of course, it's not failsafe but I'm not sure what advantage the struct has)

alexcrichton:
Ah I just opted for an opaque struct for help in terms of backwards compatibility. It looks like there are a number of standards for the tar format and they all tend to add various new types that are recognized, so I wouldn't be confident that we could exhaustively enumerate the set of types, so in that sense I'm not sure there *can* be an exhaustive match. It is true, though, that this makes it a little harder to "support everything by default" as it's easy to forget a case or two.

On the flip side, however, it provides a graceful way to handle unknown file types (e.g. weird bytes) and to get/create via raw bytes as well.

aidanhs:
Would something like
```
enum EntryType {
    Regular,
    Link,
[...]
    Unknown(u8),
}
```
allow exhaustive matching and handle strange bytes?

Yes, the one thing I do like about the struct is that the byte is stored with it, rather than having to do a lookup.

alexcrichton:
In theory, yeah, but it doesn't allow backwards-compatible expansion to new `EntryType` instances in the future (as it'd break exhaustive matches)

mbrt:
@alexcrichton I think you are right in this particular case. However I'd like to know what's the choice made in the standard library for similar cases. For example [`std::fs::Metadata`](http://doc.rust-lang.org/std/fs/struct.Metadata.html) made it the same. Is it standard to avoid enumerations when extensions are expected to come in future? Because it lessen the scope for enums very much.

[...]

PR #49, opened 2015-12-30
Add entry types for pax extended headers

dbrgn:
This should handle #48.

With this PR, the pax global header is not fully supported yet, the values are simply ignored :wink: Docs can be found [here](https://www-01.ibm.com/support/knowledgecenter/SSLTBW_1.13.0/com.ibm.zos.r13.bpxa500/pxarchfm.htm). But at least it can now unpack files that contain these headers, instead of failing.

Did you ever think about using proper enums to implement the `EntryType`s instead of using the boolean methods?

### DIFF
--- a/src/entry_type.rs
+++ b/src/entry_type.rs
@@ -1,4 +1,8 @@
 // See https://en.wikipedia.org/wiki/Tar_%28computing%29#UStar_format
+// and
+// https://www-01.ibm.com/support/knowledgecenter/SSLTBW_1.13.0/
+// com.ibm.zos.r13.bpxa500/pxarchfm.htm
+
 /// Indicate for the type of file described by a header.
 ///
 /// Each `Header` has an `entry_type` method returning an instance of this type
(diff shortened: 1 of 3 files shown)


aidanhs:
Regarding the enum, see the discussion at https://github.com/alexcrichton/tar-rs/issues/39#issuecomment-159057059

src/entry_type.rs
```diff
@@ -1,4 +1,8 @@
 // See https://en.wikipedia.org/wiki/Tar_%28computing%29#UStar_format
+// and
+// https://www-01.ibm.com/support/knowledgecenter/SSLTBW_1.13.0/
+// com.ibm.zos.r13.bpxa500/pxarchfm.htm
+
```
aidanhs: I propose linking to http://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html instead as a more accessible (and complete) page.

aidanhs:
I personally don't agree with (part of) this PR. The changes to recognise 'g' and 'x' are obviously fine, but silently ignoring a posix archive seems dangerous - tar-rs can't (currently) handle posix archives and that should be made very clear to a library consumer with an error.

If you want to unpack an archive and ignore posix headers, I would suggest iterating over the entries of an archive, checking for a posix header and ignoring it if it's a match, or unpacking otherwise. i.e. move

```
        } else if kind.is_extended_header() || kind.is_global_extended_header() {
            // Not yet implemented
            return Ok(())
```
into your own code.

src/entry_type.rs
dbrgn: Good find!

aidanhs: It's the reference I've been using :)

dbrgn:
> I personally don't agree with (part of) this PR. The changes to recognise 'g' and 'x' are obviously fine, but silently ignoring a posix archive seems dangerous - tar-rs can't (currently) handle posix archives and that should be made very clear to a library consumer with an error.

Yep. I only meant this PR as a first step in supporting PAX.

It looks like now @alexcrichton is working on adding PAX support, right?

alexcrichton:
Ok, I've opened up https://github.com/alexcrichton/tar-rs/pull/50 which now has pax support, please let me know what you think!

PR #50, opened 2016-01-20
Rewrite API and add support for GNU and pax extensions

### LINKED ISSUE #44: Long paths are truncated
Here is a tar file with a long path: [long_path.zip](https://github.com/alexcrichton/tar-rs/files/68313/long_path.zip) (zipped because github don't like tars).

Listing the content with `tar -tf long_path.tar` gives:
```
home/
home/michele/
home/michele/Documenti/
home/michele/Documenti/Development/
home/michele/Documenti/Development/Progetti/
home/michele/Documenti/Development/Progetti/MetaCloudExperiment/
home/michele/Documenti/Development/Progetti/MetaCloudExperiment/Reference/
home/michele/Documenti/Development/Progetti/MetaCloudExperiment/Reference/duplicati/
home/michele/Documenti/Development/Progetti/MetaCloudExperiment/Reference/duplicati/BuildTools/
home/michele/Documenti/Development/Progetti/MetaCloudExperiment/Reference/duplicati/BuildTools/WixIncludeMake/
home/michele/Documenti/Development/Progetti/MetaCloudExperiment/Reference/duplicati/BuildTools/WixIncludeMake/Program.cs
```

Listing the contents with this snippet:
```rust
let file = File::open("long_path.tar").unwrap();
let mut tar = Archive::new(file);
for file in tar.files_mut().unwrap() {
    if let Ok(f) = file {
        println!("{:?}", f.header().path().unwrap())
    }
}
```

Gives the following result:
```
"home/michele/"
"home/michele/Documenti/"
"home/michele/Documenti/Development/"
"home/michele/Documenti/Development/Progetti/"
"home/michele/Documenti/Development/Progetti/MetaCloudExperiment/"
"home/michele/Documenti/Development/Progetti/MetaCloudExperiment/Reference/"
"home/michele/Documenti/Development/Progetti/MetaCloudExperiment/Reference/duplicati/"
"home/michele/Documenti/Development/Progetti/MetaCloudExperiment/Reference/duplicati/BuildTools/"
"././@LongLink"
"home/michele/Documenti/Development/Progetti/MetaCloudExperiment/Reference/duplicati/BuildTools/WixIn"
"././@LongLink"
"home/michele/Documenti/Development/Progetti/MetaCloudExperiment/Reference/duplicati/BuildTools/WixIn"
```

Truncating paths and adding those `././@LongLink` entries.

### LINKED ISSUE #48: Handle pax_global_header

alexcrichton:
This is a rewrite of the API of the library (as it's a bit of a time coming at this point) as well as an inclusion of the pax and GNU extensions support. All implementations are written against [this specification](http://www.freebsd.org/cgi/man.cgi?query=tar&sektion=5&manpath=FreeBSD+8-current)

A summary of the API changes are:

* A `Builder` type has been split out of `Archive`, and now `Archive` is only meant for reading.
* The special API for `Archive` if the underlying stream is `Seek` has been removed
* The `Header` type is now by default opaque with accessors to view it as each of the particular kind of raw header that exists.
* Old type aliases were removed
* The `EntriesMut` iterator was removed as `Entries` now does the same thing (there's no special implementation based on `Seek`).
* The `Entry` type is now enhanced with a few top-level accessors to acquire data from either the extensions or the headers itself (such as path, link name, etc).

Closes #44 
Closes #48


alexcrichton:
Note that I would plan on releasing 0.4.0 once this is merged.

src/header.rs
```diff
@@ -16,47 +16,210 @@ use other;
...
+    pub fn as_old_mut(&mut self) -> &mut OldHeader {
+        self.cast_mut()
+    }
+
+    /// Flags this header as an "old header"
+    ///
+    /// Note that this is a destructive operation and may corrupt fields that
+    /// have been previously set. It is recommended to call this operation
+    /// **first** when creating new headers.
+    pub fn set_old(&mut self) {
```
mbrt: If the operation needs to be done at first, why not make a new constructor like `new_old`? And remove this function?

src/archive.rs
```diff
@@ -291,24 +182,53 @@ impl<'a> Archive<Read + 'a> {
...
     }
+}
+
+impl<'a, R: Read> Entries<'a, R> {
+    /// Indicates whether this iterator will return raw entries or not.
+    ///
+    /// If the raw list of entries are returned, then no preprocessing happens
+    /// on account of this library, for example taking into accout GNU long name
+    /// or long link archive members. Raw iteration is disabled by default.
+    pub fn raw(self, raw: bool) -> Entries<'a, R> {
```
mbrt: This method is misleading in my opinion. I'd have called it `into_<something>` as it takes `self` by value, and maybe split it in two functions like `into_raw` and `into_auto`. Another possibility is to use a function like `pub fn set_raw(&mut self, raw: bool)`.

Another thing: the documentation says "Indicates whether this iterator will return raw entries or not", but it rather turns the iterator into something else. Only a thought here. The API itself is good to have!

mbrt:
Huge improvement here! Thanks for the effort. I made some remarks, but the overall work is pretty good for me!

src/archive.rs
```diff
@@ -316,234 +236,113 @@ impl<'a> Archive<Read + 'a> {
...
+                                      but no future member found"))
+                }
+                None => return Ok(None),
+            };
+
+            if entry.header().as_gnu().is_some() &&
+               entry.header().entry_type().is_gnu_longname() {
+                if gnu_longname.is_some() {
+                    return Err(other("two long name entries describing \
+                                      the same member"))
```
aidanhs: Don't think this is actually a problem? Just overwrite the existing longname.

src/header.rs
alexcrichton: Hm true, I think I'll stick to that for now.

src/archive.rs
alexcrichton: I felt that this was somewhat different because with iterators you typically always consume by-value and don't have the `into_` prefix, but this is indeed also blending in the builder pattern a bit. I wanted to avoid a mutable method as it seems to work out best if iterators are stateless beyond the "cursor they're maintaining" (e.g. as little mutability as possible).

src/archive.rs
alexcrichton: I'm not actually sure, I couldn't necessarily find documentation indicating one way or another. I figured it'd be good to flag and be explicit, though, to prevent silently ignoring any data.

dbrgn:
Very nice!

mbrt:
All good for me!

alexcrichton:
Ok, thanks for the comments everyone! I'm gonna do some more local testing and then release 0.4.0 soon.

PR #51, opened 2016-01-22
Create and expose EntryType enum

aidanhs:
A little bit like https://github.com/alexcrichton/tar-rs/issues/39#issuecomment-159092054, but using the idea from https://github.com/alexcrichton/tar-rs/issues/39#issuecomment-171052264, here's a proposal to force non-exhausive matching, therefore allowing the enum to grow without breaking code.

It uses the exact words from https://github.com/rust-lang-nursery/regex/blob/2065b416e93cb79d61133b8a9ae6b4d0cee075b6/regex-syntax/src/lib.rs#L717-L723, which should be fine as the licenses are the same.

There are two (potential) problems:

1. you now cannot round-trip from a literal `'\0'` byte to `EntryType::Regular` back to `'\0'` - it will come out as `'0'`. One solution is to add a byte field to `EntryType::Regular`, but that makes using `==` difficult. I'm not sure this is actually a problem - '\0' is nonstandard and exists for backwards compatibility and I don't know why you'd want to round-trip to it.
2. people might use `EntryType::__Nonexhaustive` (this was called `__Unknown` but I renamed it to make the intent more clear). Given the name of the enum member, the comment immediately above it and the documentation of the enum as a whole, it would take a deliberate misuse of the library to end up in this scenario I think. The `regex` library thinks it's a reasonable risk as well. 

Ping @dbrgn

### DIFF
--- a/src/entry_type.rs
+++ b/src/entry_type.rs
@@ -3,9 +3,41 @@
 ///
 /// Each `Header` has an `entry_type` method returning an instance of this type
 /// which can be used to inspect what the header is describing.
+
+/// A non-exhaustive enum representing the possible entry types
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
-pub struct EntryType {
-    byte: u8,
+pub enum EntryType {
+    /// Regular file
+    Regular,
+    /// Hard link
+    Link,
+    /// Symbolic link
+    Symlink,
+    /// Character device
+    Char,
+    /// Block device
+    Block,
+    /// Directory
+    Directory,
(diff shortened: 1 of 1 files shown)


alexcrichton:
Alright this seems pretty good to me, I'm fine with the lack of round trip and I'm not too worried about matching against a weird variant like `__Nonexhaustive`. Thanks!

aidanhs:
Great! Thanks!

dbrgn:
:tada: :balloon: 

Thanks :)

PR #52, opened 2016-01-22
More honest internal method name

aidanhs:
Sorry, I didn't get a chance to look through #50 fully so I missed this - if I was reading tar-rs I'd be pretty suprised at `read_all` not doing what it says on the tin!

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -290,7 +290,8 @@ impl<'a> EntriesFields<'a> {
                     return Err(other("two long name entries describing \
                                       the same member"))
                 }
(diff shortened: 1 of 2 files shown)


alexcrichton:
Oh the `min` on the `with_capacity` part is actually just limiting memory allocation, this helper still consumes the entire `Entry` for all the data that it contains. It's just a little more efficient if one call to `read` can be made instead of multiple

aidanhs:
That's embarrassing, sorry.

alexcrichton:
Oh no worries! Thanks for taking a look :)

PR #55, opened 2016-01-27
Correct `into` to `dst` in `unpack` documentation

### LINKED ISSUE #54: doc: mention of variable `into` for method `unpack`
In the [documentation for unpack](http://alexcrichton.com/tar-rs/tar/struct.Archive.html#method.unpack) a variable `into` is referenced but is not listed in the function nor is it [in the source](http://alexcrichton.com/tar-rs/src/tar/archive.rs.html#87): 

"This operation is relatively sensitive in that it will not write files outside of the path specified by into." 

What is `into` in reference to?

aidanhs:
Fixes #54

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -72,7 +72,7 @@ impl<R: Read> Archive<R> {
     /// path name.
     ///
     /// This operation is relatively sensitive in that it will not write files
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oops, thanks!

PR #62, opened 2016-04-28
Implement sparse files support

tailhook:
This adds a Read trait which emits all zeros on non-existing chunks of file.

Any ideas how to provide an interface where you could actually produce a sparse file on a filesystem? As far as I know there is no cross-platform way to create a sparse file, and so no way to do it with standard library. So maybe we should provide this functionality as iterator of blocks?

I've added super minimal example files, so I could test the library. Also I should add some tests for this pull request specifically.


src/header.rs
```diff
@@ -99,6 +100,15 @@ pub struct GnuSparseHeader {
     pub numbytes: [u8; 12],
 }
 
+#[repr(C)]
+#[allow(missing_docs)]
+pub struct GnuExtSparseHeader {
+    pub sparse: [GnuSparseHeader; 21],
+    pub isextended: [u8; 1],
+}
```
alexcrichton: Could this also explicitly have the 7 bytes of padding to ensure that the header is 512 bytes in length?

src/archive.rs
```diff
@@ -243,19 +246,77 @@ impl<'a> EntriesFields<'a> {
...
+                loop {
+                    let mut ext_header = [0u8; 512];
+                    try!(read_all(&mut &self.archive.inner, &mut ext_header));
+                    self.next += 512;
+                    let sparse_header: &GnuExtSparseHeader;
+                    sparse_header = unsafe { transmute(&ext_header) };
+                    for item in &sparse_header.sparse {
+                        if item.offset[0] != 0 && item.numbytes[0] != 0 {
+                            let off = try!(octal_from(&item.offset[..]));
+                            let nbytes = try!(octal_from(&item.numbytes[..]));
```
alexcrichton: Perhaps these could be exposed as methods on the `GnuSparseHeader` type?

src/header.rs
```diff
@@ -240,11 +250,28 @@ impl Header {
         }
     }
 
+    /// Returns the size of file's data this header represents.
+    ///
+    /// This is different from `Header::size` for sparse files, which have
+    /// some longer `size()` but shorter `data_size()`.
+    ///
+    /// May return an error if the field is corrupted.
+    pub fn data_size(&self) -> io::Result<u64> {
```
alexcrichton: Perhaps this could be something like `entry_size`? It's just the number of bytes following the header, right?

src/archive.rs
```diff
@@ -243,19 +246,77 @@ impl<'a> EntriesFields<'a> {
...
+                            if nbytes > 0 {
```
alexcrichton: Is this common or an old compatibility thing as well?

src/archive.rs
```diff
@@ -243,19 +246,77 @@ impl<'a> EntriesFields<'a> {
...
+                    }
+                }
+            }
+            if gnu.isextended[0] == 1 {
+                loop {
+                    let mut ext_header = [0u8; 512];
+                    try!(read_all(&mut &self.archive.inner, &mut ext_header));
+                    self.next += 512;
+                    let sparse_header: &GnuExtSparseHeader;
+                    sparse_header = unsafe { transmute(&ext_header) };
```
alexcrichton: I'd prefer if this had a similar `as_mut_bytes` method like the other header entries to make this a little safer

src/reader.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// if you feed the file to some streaming parser or whatever.
+    fn read(&mut self, mut into: &mut [u8]) -> io::Result<usize> {
+        match *self {
+            Reader::Normal(ref mut reader) => reader.read(into),
+            Reader::Sparse {
+                ref mut data, ref blocks, ref mut block,
+                ref mut position, size,
+            } => {
+                let mut bytes_read: usize = 0;
+                while into.len() > bytes_read {
```
alexcrichton: This seems like it's a pretty complicated loop, perhaps it'd be better expressed as an iterator?

Given an iterator of chunks, it seems like you could `map` that to an iterator of I/O objects like `io::repeat` or the actual underlying archive (alternating).

Perhaps that could simplify some of this?

src/archive.rs
```diff
@@ -243,19 +246,77 @@ impl<'a> EntriesFields<'a> {
...
+                    }
+                    bytes += chunk.1;
+                    offset = chunk.0 + chunk.1;
+                }
+                if bytes != data_size {
+                    return Err(other("wrong total length of data for sparse file"))
+                }
+                if offset > size {
+                    return Err(other("sparse chunk end exceeds file size"))
+                }
```
alexcrichton: It might be best to perform this validation as we're reading the chunks above as well, may help some ddos like situations as well perhaps

alexcrichton:
Awesome, thanks @tailhook! Could you also be sure to add enough tests to cover all the new code here as well?

src/archive.rs
tailhook: Do you ask about checking nbytes? I think this is just a validation issue. We don't need empty blocks regardless of whether they are allowed by spec or just carefully put into into a malicious tar.

src/archive.rs
tailhook: Well, this is a very simple loop. We have already gathered all the chunks into a Vec, visiting them once again is cheaper than that. Doing validation later is unclear: when should we do it? On first `read()`? Then we'll pay on each next read a price of checking a boolean flag.

tailhook:
I'm don't know how to make it simpler or how to write decoder in terms of `io` combinators. We can't do chain of the arbitrary number of readers not known at compile time. And even if we do, we can't borrow the reader twice.

I've fixed other issues and added tests. So given we now have the tests, can we merge it in, and refactor later if needed?

alexcrichton:
Ok, I ended up rejiggering the code a bit and adding a few more pieces of functionality here and there, can you give it a whirl and see if it works for you?

tailhook:
Works fine for me. Thanks!

PR #66, opened 2016-06-12
Implement set_unpack_xattrs() functions

noiseless:
First commit adds two new functions:
```rust    
pub fn Archive::unpack_preserving_xattrs<P: AsRef<Path>>(&mut self, dst: P) -> io::Result<()>
pub fn Entry::unpack_preserving_xattrs<P: AsRef<Path>>(&mut self, dst: P) -> io::Result<()>
```   
This functions have a similar behavior as unpack(), but preserves extended
file attributes. That also means, that this function shouldn't works if dst
is placed in pseudo file systems.

I decided to use a separate function due to the fact that adding support for extended
attributes makes it impossible to extract data in tmpfs or other pseudo-fs.
This limits the scope of function; besides gnu tar, for example, also requires
a separate argument for preserving xattrs. Adding a flag to the unpack() would
break backward compatibility.

Second commit adds two simple tests for these functions


noiseless:
What should I write in cargo.toml to disable xatrr on Windows?

alexcrichton:
Thanks! You can have a platform-specific dependency like:

```toml
[target.'cfg(unix)'.dependencies]
xattr = "..."
```

src/entry.rs
```diff
@@ -309,6 +320,29 @@ impl<'a> EntryFields<'a> {
                                         for `{}`", mode, dst.display()), e)
             }));
         }
+        // Windows does not completely support posix xattrs
+        // https://en.wikipedia.org/wiki/Extended_file_attributes#Windows_NT
+        if cfg!(target_family = "unix") && preserve_xattrs {
```
alexcrichton: Is there a downside from doing this by default? May be worth seeing if `tar` does that by default

noiseless: By default, gnu tar does not preserve extended attributes. You need to specify "--xattrs" option.
If we want to preserve xattrs by default, we probably need to ignore any
errors while setting extended attributes, if we do not want to prohibit the use of unpack() on tmpfs or other virtual fs*.

Should I delete functions unpack_preserving_xattrs() and move processing xattrs in unpack(), ignoring xattrs::set() errors?

* As I understand from the documentation, linux does not support user_xattr for tmpfs in vanilla kernel. Only  ACLs/capabilities/other system attributes. Perhaps, ignoring errors of setting these attributes is the right decision.

alexcrichton: Ah ok, then in that case we should probably follow the same behavior as `tar` and not do it by default. I'd personally prefer to avoid `unpack` and `unpack_foo` method duplication, so perhaps an option could be set somewhere? Like maybe `Archive::set_unpack_xattrs` which just sets a boolean flag.

We'd probably also want it to be a unix-specific extension trait, but I can take care of that part as well.

noiseless:
> Ah ok, then in that case we should probably follow the same behavior as tar and not do it by default. I'd personally prefer to avoid unpack and unpack_foo method duplication, so perhaps an option could be set somewhere? Like maybe Archive::set_unpack_xattrs which just sets a boolean flag.
> 
> We'd probably also want it to be a unix-specific extension trait, but I can take care of that part as well.


Thanks for the clarification!
Corrected code according to your suggestions.

alexcrichton:
Oh oops I rebased so it didn't merge automatically, but I've now pushed this to master!

noiseless:
Thank you!

PR #69, opened 2016-06-19
Fix Adding Hidden Directories

mmstick:
This change will check if there is one instance of `Component::CurDir`,
and if so, if a path is "." then a "." directory will be created.


alexcrichton:
Thanks! Can you add a test for this, and also add a test for when adding a path like `foo/./bar` it comes back with a path name of `foo/bar`?

Issue #70, opened 2016-07-13
7z shows extra directories in tar-rs archive


lnicola:
I created an archive with two files using `append_path`. Running `7z l` gives me:

```
Type = tar
Physical Size = 27136
Headers Size = 2048
Code Page = UTF-8

   Date      Time    Attr         Size   Compressed  Name
------------------- ----- ------------ ------------  ------------------------
2016-07-12 23:59:56 .....          232          512  12741255114/Cargo.toml
2016-07-13 00:00:10 .....        24496        24576  12740666100/Cargo.lock
------------------- ----- ------------ ------------  ------------------------
2016-07-13 00:00:10              24728        25088  2 files
```

Doing the same on an archive created with `tar` (ignore the different entry order):

```
Type = tar
Physical Size = 30720
Headers Size = 5632
Code Page = UTF-8

   Date      Time    Attr         Size   Compressed  Name
------------------- ----- ------------ ------------  ------------------------
2016-07-13 00:00:10 .....        24496        24576  Cargo.lock
2016-07-12 23:59:56 .....          232          512  Cargo.toml
------------------- ----- ------------ ------------  ------------------------
2016-07-13 00:00:10              24728        25088  2 files
```

`tar` extracts the archive just fine, but it can be a hassle for `7z` users, especially on Windows. I haven't tried other archiving programs.

alexcrichton:
Oh dear, sounds bad! Could you upload the code you use to create the archive, along with the two archives in question?

alexcrichton:
It looks like maybe the timestamp is leaking into the filename? Seems odd...

lnicola:
Oh, I just used the example on the GitHub page.

lnicola:
```
hexdump -C foo.tar # created with tar-rs
00000000  66 69 6c 65 31 2e 74 78  74 00 00 00 00 00 00 00  |file1.txt.......|
00000010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000060  00 00 00 00 30 30 30 30  36 34 34 00 30 30 30 31  |....0000644.0001|
00000070  37 35 30 00 30 30 30 31  37 35 30 00 30 30 30 30  |750.0001750.0000|
00000080  30 30 30 30 30 30 32 00  31 32 37 34 31 33 35 30  |0000002.12741350|
00000090  35 36 37 00 30 30 31 33  31 36 33 20 30 00 00 00  |567.0013163 0...|
[...]

PR #71, opened 2016-07-14
Don't fill atime and ctime (7-zip compat)

lnicola:
The atime and ctime fields in the GNU entry header seem to be
deprecated. GNU tar only fills them in when --incremental is passed and
the manpage calls that variation "old GNU-format".

Filling the two fields is problematic because 7-zip (which is rather
popular on Windows) interprets the timestamps as directories. WinRAR is
not affected, but libarchive ignores the two fields. Since they are not
well-supported, it's better to leave them empty.

Fixes: #70


alexcrichton:
Thanks!

PR #77, opened 2016-07-23
Use the right name for @LongLink entries

### LINKED ISSUE #76: GNULongLink entries have the wrong name
`././@LongLink` seems to be the standard name, but `tar-rs` emits them as `@LongLink`. I don't have any example of an issue caused by this, however, I expect fixing it will improve compatibility with other applications and libraries.

lnicola:
These seem to be found with a name of `././@LongLink`, however
`copy_path_into` emits that as `@LongLink`. This is more tricky to fix
because `Path::components()` collapses that into `./@LongLink`. For now,
just special-case them.

The implementation here is not ideal since it will change the behavior
for non-GNU entries.

Fixes #76

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -243,7 +243,7 @@ fn append_fs(dst: &mut Write,
             return Err(e)
         }
         let mut header2 = Header::new_gnu();
-        try!(header2.set_path("././@LongLink"));
+        header2.as_gnu_mut().unwrap().name[..13].clone_from_slice(b"././@LongLink");
         header2.set_mode(0o644);
         header2.set_uid(0);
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! I wonder if though this could just directly set the bytes of the header [here](https://github.com/alexcrichton/tar-rs/blob/master/src/builder.rs#L246) with `.as_gnu().unwrap().name`?

lnicola:
Hmm. I can try that tomorrow.

lnicola:
Is this the right way to copy elements from one array to another? :worried:

alexcrichton:
Fine by me, thanks!

PR #79, opened 2016-07-23
Fix clippy warnings

lnicola:


src/archive.rs
```diff
@@ -420,7 +420,7 @@ impl<'a> EntriesFields<'a> {
                 while ext.is_extended() {
                     try!(read_all(&mut &self.archive.inner, ext.as_mut_bytes()));
                     self.next += 512;
-                    for block in ext.sparse.iter() {
+                    for block in &ext.sparse {
```
alexcrichton: I personally prefer `.iter()` instead of `&foo`, could this change back?

PR #85, opened 2016-11-23
Add an Entry::unpack_in to do safe unpacking of one Entry

joshtriplett:
This makes it easy to walk over each Entry in an archive, filter them in
some way, and unpack some subset of them, without reimplementing the
security-sensitive logic in Archive::unpack.


alexcrichton:
Thanks!

PR #90, opened 2016-12-29
entry: relax symlink target constraints

### LINKED ISSUE #87: Fix for symlink attack needs to support extracting symlinks pointing outside the root
Thanks for fixing the symlink attack.  However, the fix in 85b6399b212f66eb955431f214e0e9eb408988c8, while addressing the vulnerability, also prevents extracting valid archives that contain a symlink pointing outside the root, even if nothing attempts to write "through" those symlinks.  As an example of such a valid tarball, consider extracting the filesystem tarball from a Debian package into a temporary directory.  Such a tarball could legitimately contain, for instance, a symlink like `/usr/bin/vi -> /etc/alternatives/vi`; that shouldn't cause extracting the archive to a temporary directory to fail.  It's even OK to have a symlink to a directory outside the root, as long as nothing gets extracted *through* that symlink.

The fix for busybox (in https://bugs.busybox.net/show_bug.cgi?id=8411) used the approach of writing a "placeholder" for such symlinks, and replacing all of the placeholders with symlinks at the end of the extraction; then, if anything attempts to extract "through" that placeholder, it'll fail.

lucab:
This allows the creation and extraction of symlinks pointing to targets
outside of the destination directory, including absolute paths and `../`
chains.

Path traversal is now checked at unpack time, by canonicalizing the target
path and ensuring that it is nested below destination path.

Fixes #87

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -2,9 +2,9 @@ use std::borrow::Cow;
 use std::cmp;
 use std::fs;
 use std::io::prelude::*;
-use std::io::{self, SeekFrom};
+use std::io::{self, Error, ErrorKind, SeekFrom};
 use std::marker;
-use std::path::{Component, Path, PathBuf};
+use std::path::{Component, Path};
 
 use filetime::{self, FileTime};
 
@@ -155,7 +155,7 @@ impl<'a, R: Read> Entry<'a, R> {
     /// }
     /// ```
     pub fn unpack<P: AsRef<Path>>(&mut self, dst: P) -> io::Result<()> {
-        self.fields.unpack(dst.as_ref(), None)
+        self.fields.unpack(dst.as_ref())
     }
 
     /// Extracts this file under the specified path, avoiding security issues.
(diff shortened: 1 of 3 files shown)


src/entry.rs
```diff
@@ -358,54 +372,15 @@ impl<'a> EntryFields<'a> {
...
-                    }
-                }
-            }
-            if actual_src.iter().count() == 0 {
+            if src.iter().count() == 0 {
                 return Err(other("symlink destination is empty"))
             }
 
-            println!("{:?} {:?}", actual_src, dst);
+            println!("{:?} {:?}", src, dst);
```
alexcrichton: oh oops this is actually a stray println and should get removed!

tests/entry.rs
```diff
@@ -130,5 +180,8 @@ fn good_parent_paths_ok() {
 
     let td = t!(TempDir::new("tar"));
     t!(ar.unpack(td.path()));
-    t!(File::open(td.path().join("foo/bar")));
+    t!(td.path().join("foo").join("bar").read_link());
+    let dst = t!(td.path().join("foo").join("bar").canonicalize());
```
alexcrichton: It looks like this line is failing on AppVeyor?

alexcrichton:
Looks great to me, thanks! I'm fine merging once AppVeyor is green

src/entry.rs
lucab: Yes, I re-introduced it later in order to get some clues about the appveyor issue.

tests/entry.rs
lucab: Yes, and I'm puzzled. I expanded the previous test into path-joining, readlink(), and canonicalize() in order to understand what's going on, but the original one was also failing. It looks like windows doesn't like the `../bar` symlink. This crate is converting `\` to `/` in symlinks, so `..\bar` becomes `../bar`. This may be part of the issue, but it seems to be on purpose (and actually covered by unit testing). I'm not familiar with windows api, how do relative symlinks and separators work there?

alexcrichton: Hm to be honest I'm surprised that symlinks work at all on Windows, they're supposed to be gated by default...

Want to just `#[cfg(unix)]` the symlink tests?

lucab:
As suggested, moved the last test to unix-only. PTAL.

alexcrichton:
Thanks!

PR #91, opened 2016-12-29
Make cast functions safe to use

Manishearth:
As it stands it's basically a safe transmute. Add a bound so that you can only transmute to known header formats.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -13,6 +13,13 @@ use std::str;
 use EntryType;
 use other;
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
This feels unnecessary? This is an internal function that's only called in a few places. If we really care why not just make it unsafe and put unsafe at the callsites?

Manishearth:
https://github.com/alexcrichton/tar-rs/pull/93

Issue #89, opened 2016-12-09
Support adding a directory recursively


joshtriplett:
I'd love to have a Builder method to add a directory recursively, to avoid having to walk it by hand.

gyscos:
Seems implemented with [`Builder::append_dir_all`](https://docs.rs/tar/0.4.13/tar/struct.Builder.html#method.append_dir_all).

alexcrichton:
Indeed!

PR #96, opened 2017-02-12
Add a Builder::append_dir_all method to add a directory recursively

mdsteele:
Adds an `append_dir_all` method (named by analogy to e.g. `fs::remove_dir_all`) to `Builder` that recursively adds a directory and all of its contents to the archive (see issue #89).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ contents are never required to be entirely resident in memory all at once.
 [dependencies]
 libc = "0.2"
 filetime = "0.1.5"
+walkdir = "1.0"
 
 [dev-dependencies]
 tempdir = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ contents are never required to be entirely resident in memory all at once.
 [dependencies]
 libc = "0.2"
 filetime = "0.1.5"
-walkdir = "1.0"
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for the PR! Could this perhaps be implemented without `WalkDir` though? I think it'd be pretty simple to write a simple recursive walking function as well to visit dirs/files.

mdsteele:
Sure thing; here's take 2, this time with 100% less `WalkDir`.

alexcrichton:
Looks great!

As a minor nit, I think this'll allow being called with a path to a file due to the embedded check for whether it's a file or not, perhaps we could restrict this for now to only be called with directories by initially calling `read_dir`?

Also, the `entry` coming out of `read_dir` has a `file_type` method which should quickly allow us to learn about file/directory distinctions, perhaps we could leverage that to avoid extra syscalls here too?

mdsteele:
Done and done.  The `append_dir_all` method now uses the file type from the `DirEntry` to save on syscalls, and I've arranged it so that the first operation performed on the initial base path will always be a `read_dir`, which will fail as desired if the path isn't a directory.  I've also added a new test to check that `append_dir_all` does in fact fail when called on a regular file.

alexcrichton:
Looks great, thanks!

PR #98, opened 2017-02-15
Query raw byte positions of the header and file of an entry

peterhj:
Implement methods on `Entry` for querying the raw byte positions of the entry's header and file within the archive.

The main motivation is that there are cases when the `Read` object passed to `Archive` implements `Seek` or is simply byte-sliceable, and we want to perform random access on individual entries of the archive. A concrete example is a large memory mapped tar file, for which we want to avoid unnecessary copies through `Read`.

This commit does what seems to be the least invasive thing to achieve the above: (1) adding two helper methods to `Entry` for querying the raw byte positions of the header and file, and (2) putting the responsibility to the user for figuring out whether it is valid to use the raw byte representations (e.g. when the entry type is not a sparse file).

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -183,6 +183,7 @@ impl<'a> EntriesFields<'a> {
         let delta = self.next - self.archive.inner.pos.get();
         try!(self.archive.skip(delta));
 
+        let header_pos = self.next;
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks great to me, thanks!

Issue #97, opened 2017-02-12
Expose Gnu LongName header extension


stuhood:
`Builder::append_file` and `Builder::append_file` implement support for `LongName` via `builder::append_fs`, which writes two Headers if necessary to store the filename. But (understandably) `Builder::append` does not.

Users who need to strip or manipulate the file metadata while taring something (in particular, because they're looking for reproducible, byte-identical output) might use the `Builder::append` method with a `Header` containing only normalized metadata; but unfortunately, this will not trigger the `LongName` extension, meaning that that feature would need to be implemented by the caller.

----

I don't really see a really obvious way to expose the `LongName` logic with the current API, since the caller will fail in `Header::set_path`, and the extension requires not just another Header but also some content in the entry.

One option might be to add a synthetic Header type (ie, one that is not exactly 512 bytes in length) that will accept all data that can possibly be appended when extensions are in play: ie, has an unbounded `path` length. Then, during `append` for this Header type, usage of extensions would be pre-acknowledged.

stuhood:
Happy to implement this if there is a reasonable approach!

alexcrichton:
Thanks for the issue! I wonder if we could perhaps tackle this through two separate avenues:

1. Add a `deterministic` flag to `Builder`. When set it could ensure that deterministic archives were ignoring, ignoring metadata such as mtime/atime/etc.
2. Add a new method, `append_name` (or something like that) which takes a `&Path`, a `&mut Header`, and a `&mut Read`. That'd auto add a LongName header ahead of time if necessary and otherwise just put the `Path` into the `Header` provided. 

How's that sound?

stuhood:
Thanks @alexcrichton ! Either of those would be reasonable. `(1)` sounds like the most general option... `(2)` is a tiny fraction better for my usecase (we can use information from previous syscalls to populate the size, so `append_fs`'s lookup of the metadata is extraneous).

Calling my "synthetic header type" suggestion option `(3)`, I'd wonder whether there are any other extensions that might benefit from looser constraints at header construction time... if so, then continuing to overload the `append*` methods (ie, `(2)`) could get hairy.

alexcrichton:
Yeah it's true that there's turning out to be quite a few append functions... 

[...]

PR #99, opened 2017-02-18
Add a deterministic append mode

stuhood:
As discussed on #97 , this adds a deterministic append mode via a new `HeaderMode` enum that can be set on a `Builder`.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -5,22 +5,25 @@ use std::fs;
 use std::borrow::Cow;
 
 use {EntryType, Header, other};
-use header::{bytes2path, path2bytes};
+use header::{bytes2path, HeaderMode, path2bytes};
 
 /// A structure for building archives
 ///
 /// This structure has methods for building up an archive from scratch into any
 /// arbitrary writer.
 pub struct Builder<W: Write> {
+    mode: HeaderMode,
     finished: bool,
     obj: Option<W>,
 }
 
 impl<W: Write> Builder<W> {
     /// Create a new archive builder with the underlying object as the
-    /// destination of all data written.
+    /// destination of all data written. The builder will use
+    /// `HeaderMode::Complete` by default.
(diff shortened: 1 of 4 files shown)


src/builder.rs
```diff
@@ -30,6 +33,13 @@ impl<W: Write> Builder<W> {
         self.obj.as_mut().unwrap()
     }
 
+    /// Changes the HeaderMode that will be used when reading fs Metadata for
+    /// methods that implicitly read metadata for an input Path. Notably, this
+    /// does _not_ apply to `append(Header)`.
+    pub fn mode(&mut self, mode: HeaderMode) {
+        self.mode = mode;
```
stuhood: Wanted to follow the pattern set with `Entries.raw` here, but was informed that because `Builder` implements `Drop`, that's not possible.

alexcrichton:
Looks great to me, thanks!

Could you add a `__Nonexhaustive` variant documented as "please don't match this" to the enum as well? Other than that r+ from me

stuhood:
@alexcrichton : Would you mind pointing me to a good example of doing that? Would the variant be marked `#unstable`? If so, what would the syntax be?

alexcrichton:
Oh yeah sure I'm just thinking of [something like this](https://github.com/alexcrichton/toml-rs/blob/3b7494cf501c6e18b249977c7f0d56654c5e381c/src/ser.rs#L97-L98). It basically just discourages exhaustive matches so we can save ourselves to add more variants in the future (if need be)

stuhood:
Ok, added `__Nonexhaustive`.

alexcrichton:
Looks like CI may be failing?

stuhood:
@alexcrichton : Yea, sorry... needed a merge from master.

alexcrichton:
Awesome, thanks!

PR #104, opened 2017-04-24
Make append_dir_all() accept "" as dest

### LINKED ISSUE #103: Support "." as first argument of append_dir_all()
The command `builder.append_dir_all(".", "foo")` fails with the error "paths in archives must have at least one component".

It seems like it would be useful to support this, to populate the root of the archive from the contents of a folder. So `foo/bar` would be in the archive as `./bar`, etc. Does this seem like a good thing to add, or is there a better way to do this?

Implementing this should just be a matter of not calling `append_dir()` for the top level directory if it is `.`.

ids1024:
Fills root of archive from that directory. Closes #103.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -316,7 +316,9 @@ fn append_dir_all(dst: &mut Write, path: &Path, src_path: &Path, mode: HeaderMod
                 let entry = try!(entry);
                 stack.push((entry.path(), try!(entry.file_type()).is_dir()));
             }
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Could you add a test for this as well?

ids1024:
Ok; I've added a test, which is largely the same as the primary test for `append_dir_all()`.

alexcrichton:
Thanks!

PR #106, opened 2017-05-11
Normalize all path separators to Unix-style `/`

cuviper:
In most cases, `tar` already wrote only `/` separators between path
components.  This is the only separator that works on Unix, and it's
acceptable on Windows too.  However, long paths in GNU headers were
writing the full `path2bytes` result without translation, which meant
Windows could write tar files with `\` that aren't portable to Unix.

Now `path2bytes` will normalize all `\` to `/` on Windows.

The new `path_separators` test makes sure that short and long paths with
UStar and GNU headers will round-trip correctly, meaning the `path()`
compares equal, while checking that `path_bytes()` doesn't contain `\`
path separators.  The long GNU path was writing `\` on Windows before.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -1031,7 +1031,7 @@ fn copy_path_into(mut slot: &mut [u8],
                 return Err(other("path component in archive cannot contain `/`"))
             }
         }
-        try!(copy(&mut slot, bytes));
+        try!(copy(&mut slot, &*bytes));
         emitted = true;
     }
     if !emitted {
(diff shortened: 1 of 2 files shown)


cuviper:
FWIW, I did find 1382ad45add193de43fee56add607c632f54dcbc which tweaked this behavior before, but that appears to be more about how paths were *read*.  And it sounds like the intention was that Windows would still translate to `/`, but the GNU long names appear to be a buggy loophole.  So here, I'm only translating how paths are converted *into* the tar file, and only on Windows.  If a Unix user wants to write a filename with `\`, they still can.

alexcrichton:
Everything here looks spot on, thanks so much @cuviper! I'll publish this with a new release.

alexcrichton:
Note that this doesn't actually affect cargo, ... yet. For weird historical reasons Cargo still uses ustar archives, but I think it's been ~years since that bug was fixed so this'll be a very welcome fix when it switches to gnu archives!

PR #107, opened 2017-05-22
Include file mode in items filtered for deterministic headers

stuhood:
For our usecase of the `Deterministic` header mode, we're not interested in including anything except the `execute` bit of appended files (in particular, we're trying to align with what `git` persists for files). See https://github.com/pantsbuild/pants/issues/4610 for more color.

The change introduces a mask that drops all filemode information other than the type and the execute bit. It's possible that this could be considered a new Header mode, but I'd prefer to just change the behaviour of the existing `Deterministic` mode if that works for you.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -636,7 +636,7 @@ impl Header {
                 self.set_uid(0);
                 self.set_gid(0);
 
-                // Use a default umask value, but propogate the (user) execute bit.
+                // Use a default umask value, but propagate the (user) execute bit.
                 let fs_mode =
                   if meta.is_dir() || (0o100 & meta.mode() == 0o100) {
                     0o755
--- a/src/header.rs
+++ b/src/header.rs
@@ -10,6 +10,8 @@ use std::mem;
 use std::path::{Path, PathBuf, Component};
 use std::str;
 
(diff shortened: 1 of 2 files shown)


src/header.rs
```diff
@@ -131,6 +133,9 @@ pub struct GnuExtSparseHeader {
     pub padding: [u8; 7],
 }
 
+/// Mask to select only file type and the executable bit.
+const DETERMINISTIC_FILE_MODE_MASK: u32 = libc::S_IFMT as u32 | libc::S_IXUSR as u32 | libc::S_IXGRP as u32 | libc::S_IXOTH as u32;
```
alexcrichton: I personally always find these a little wonky to read, but the octal notation should work well here, right?

alexcrichton: Also, it looks like these may not be defined on Windows?

alexcrichton:
Sounds great to me! Thanks for the PR!

stuhood:
In testing it became clear that not all tar applications apply a default umask during extraction, so having a default umask during creation is helpful. Applied that change, as well as switching to octal.

alexcrichton:
Looks great!

PR #110, opened 2017-05-30
Add `Builder::append_data` supporting long paths

### LINKED ISSUE #108: Long filenames when adding via append<R: Read>
It would be great if the logic that automatically [prepends a long filename header](https://github.com/alexcrichton/tar-rs/blob/1647ce76ee09cfed3fb0e909517ac7e65e3a1f6c/src/builder.rs#L282-L303) was also available for the `append` method. Currently the code is not reusable and has to be copy-pasted if one wants to pass in `Read` implementations associated with long file names as opposed to `Path`.

### LINKED ISSUE #109: Need a way to add executable Windows files
I'm looking how to whitelist Windows files that should be marked executable, as discussed in rust-lang/rust#42121, but I don't see a clean way to do this.  The easy `Builder::append_path()` and `append_file()` methods call `append_fs()` which fills the header mode from file metadata, and the Windows `fill_platform_from()` only considers whether files are read-only.  (There's not a Windows concept of "executable" mode unless you dip into ACLs.)

AFAICS the only method we have for setting a manual mode is `Builder::append()` with your own header, but that will miss an important bit in `append_fs()` which deals with paths that are too long for a normal GNU header.  I don't think we want that logic duplicated by `tar` users.

So I think `Builder` will need a new method to give more control over the file mode, while keeping other niceties like long-path handling.

cuviper:
This method lets one specify a manual header like `append`, but
incorporates the special handling needed for setting long paths.  If
given path doesn't fit in the header `name`, it will write a GNU
`././@LongLink` entry beforehand.

Fixes #108.
Fixes #109.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -93,6 +93,53 @@ impl<W: Write> Builder<W> {
         append(self.inner(), header, &mut data)
     }
 
+    /// Adds a new entry to this archive with the specified path.
+    ///
+    /// This function will set the specified path in the given header, which may
+    /// require appending a GNU long-name extension entry to the archive first.
+    /// The checksum for the header will be automatically updated via the
+    /// `set_cksum` method after setting the path. No other metadata in the
+    /// header will be modified.
+    ///
+    /// Then it will append the header, followed by contents of the stream
+    /// specified by `data`. To produce a valid archive the `size` field of
+    /// `header` must be the same as the length of the stream that's being
+    /// written.
+    ///
+    /// Note that this will not attempt to seek the archive to a valid position,
+    /// so if the archive is in the middle of a read or some other similar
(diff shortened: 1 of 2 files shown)


alexcrichton:
👍 

Looks great, thanks!

alexcrichton:
Mind testing this out upstream before I publish a new version? If it works I'll publish pronto

cuviper:
I tried it here https://github.com/cuviper/rust-installer/commit/5ad2d00703fd0ad52f9a8bef2adeaa8a6eb42976, and updated into the rust submodule, and it seems to work well.  I don't love a whitelist, but I don't have a better option. 🙂

alexcrichton:
Awesome! I've published 0.4.13

PR #113, opened 2017-06-11
entry: fix unpacking relative hardlinks

### LINKED ISSUE #112: entry/unpack: failure extracting relative hardlinks
I've encountered a failure when extracting a busybox image, which contains a single binary and many hardlinked copies. Those links are referenced by their relative path, e.g. `bin/true -> bin/busybox`.

Extracting such archive with `Archive::unpack()` results in an OS error because this library tries to use the relative path as-is, instead of nesting it under the target directory:
```
[pid  8042] link("bin/busybox", "/tmp/foo/bin/true") = -1 ENOENT (No such file or directory)
```

A short reproducer is the following testcase, currently failing on master:
```rust
#[test]
fn relative_hardlink() {
    let mut ar = tar::Builder::new(Vec::new());

    let mut header = tar::Header::new_gnu();
    header.set_size(0);
    header.set_entry_type(tar::EntryType::Regular);
    t!(header.set_path("foo"));
    header.set_cksum();
    t!(ar.append(&header, &[][..]));

    let mut header = tar::Header::new_gnu();
    header.set_size(0);
    header.set_entry_type(tar::EntryType::Link);
    t!(header.set_path("bar"));
    t!(header.set_link_name("foo"));
    header.set_cksum();
    t!(ar.append(&header, &[][..]));

    let bytes = t!(ar.into_inner());
    let mut ar = tar::Archive::new(&bytes[..]);

    let td = t!(TempDir::new("tar"));
    t!(ar.unpack(td.path()));
    t!(td.path().join("foo").metadata());
    t!(td.path().join("bar").metadata());
}
```

lucab:
This commit fixes an unpacking failure triggered by hardlinks with
relative link names. In such case, the source of the hard-link
must be resolved first by nesting it under the unpacking directory.

Fixes #112

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -177,7 +177,7 @@ impl<'a, R: Read> Entry<'a, R> {
     /// }
     /// ```
     pub fn unpack<P: AsRef<Path>>(&mut self, dst: P) -> io::Result<()> {
-        self.fields.unpack(dst.as_ref())
+        self.fields.unpack(None, dst.as_ref())
     }
 
     /// Extracts this file under the specified path, avoiding security issues.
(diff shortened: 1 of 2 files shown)


alexcrichton:
Awesome, thanks for the fix!

PR #115, opened 2017-06-18
Unwrap returned Result in documentation for Builder

### LINKED ISSUE #114: Documentation for `Builder` ignores result from `Header::set_path`
The examples for `Builder` include a few calls to `Header::set_path` but ignore its `Result`. Reading these examples, it's not clear a `Result` is returned at all.

Maybe this is intentional, since the focus is on `Builder` rather than `Headers`? Still, adding `.unwrap()` might make things clearer.

gyscos:
Fixes #114


alexcrichton:
Thanks!

PR #120, opened 2017-07-10
Redox OS support

ids1024:


ids1024:
Building on Redox also depends on https://github.com/alexcrichton/filetime/pull/8.

PR #131, opened 2017-10-03
handle extended numeric formats in octal_from

### LINKED ISSUE #130: failure "numeric field did not have utf-8 text" with large files in archive
```
cargo new example
cd example
fallocate -l 8G large.file
tar cf large.tar large.file
echo 'tar = "0.4"' >> Cargo.toml
cat << EOF > src/lib.rs
extern crate tar;

use std::fs::File;
use tar::Archive;

#[test]
fn load() {
    let archive = File::open("large.tar").unwrap();
    let mut archive = Archive::new(archive);
    for entry in archive.entries().unwrap() {
        entry.unwrap();
    }
}
EOF
cargo test
```

results in

```
running 1 test
test load ... FAILED

failures:

---- load stdout ----
        thread 'load' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Custom(Custom { kind: Other, error: StringError("numeric field did not have utf-8 text") }) }', /checkout/src/libcore/result.rs:860:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    load

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

edit: fails with 8GB file at a minimum, I had 36GB  before

cake4289:
fixes #130

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -1153,13 +1153,24 @@ impl Default for GnuExtSparseHeader {
 }
 
 fn octal_from(slice: &[u8]) -> io::Result<u64> {
-    let num = match str::from_utf8(truncate(slice)) {
-        Ok(n) => n,
-        Err(_) => return Err(other("numeric field did not have utf-8 text")),
-    };
-    match u64::from_str_radix(num.trim(), 8) {
-        Ok(n) => Ok(n),
-        Err(_) => Err(other("numeric field was not a number"))
+    if slice[0] & 0x80 != 0 { // GNU length extension
+        let mut total: u64 = (slice[0] ^ 0x80) as u64;
+        let mut index = 1;
+        while index < slice.len() {
+            total <<= 8;
+            total |= slice[index] as u64;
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! Do you know of some docs online where this could link to which explain this extension?

cake4289:
See https://www.freebsd.org/cgi/man.cgi?query=tar&sektion=5&manpath=FreeBSD+8-current under section "Numeric Extensions".

cake4289:
might want to do this in octal_into as well I guess though I haven't done that here

alexcrichton:
Ok thanks! Could you add that link in the code?

Also, perhaps we could avoid changing `octal_into`? It sounds like this isn't a super-well-supported encoding and AFAIK there weren't many problems with the previous incarnation?

cake4289:
done & done

alexcrichton:
Thanks!

Issue #119, opened 2017-07-06
Support archiving hard links


ids1024:
Tar (at least GNU tar) seems to notice when files being archived are hard links, and stores them as such, extracting as hard links.

I'm not quite sure how this should be implemented (and besides, I don't currently/yet need it) but it should have an option like the symlink option in #117.

alexcrichton:
Makes sense to me!

BrianOn99:
I looked into how busybox tar and gnu tar did it.  The way is to record in a HashSet the (inode, device number) of every file they have written, if it has hard link count > 1.  Then when they see this (inode, device number) a second time, jsut the header is written, with typeflag set to 1 and an appropiate linkname.

The appropiate place to set the type flag should be `append_fs` in `Builder`, because it is where the header is prepared.  However in order for `append_fs` to know it is a hard link, Builder has to tell it, that means adding a parameter to `append_fs`.  That would make `append_fs` having 7 parameters and it may be too many.  Every caller such as `append_path` has to pass the parameter around too.  So, I am considering to put `append_fs` inside `impl Builder`
```
fn append_fs(&mut self,
             path: &Path,
             meta: &fs::Metadata,
             read: &mut Read,
             link_name: Option<&Path>) -> io::Result<()> {
```
So that `dst`, `mode` and the new HashSet of (inode, device number) can be retreived from self.

That would be a big change as some other module-level function like `append_path` will be moved to self too.  I would like to ask if this is acceptable, or there is a better approach?

alexcrichton:
@BrianOn99 sounds great to me!

BrianOn99:
@alexcrichton Then I will take this issue, while make sure not changing external API.

the8472:
See #192, extending `append_fs` would be too high level when not building an archive from the filesystem but passing in partial headers + paths + readables directly.

PR #135, opened 2017-11-11
Archiving hard link

BrianOn99:
Fix issue #119 
I have not tested the code in redox and windows.

There is a little bit code duplication in `append_fs` but I cannot find a good solution.  Any advice would be welcome.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,11 +1,14 @@
-#[cfg(any(unix, target_os = "redox"))] use std::os::unix::prelude::*;
+#[cfg(any(unix, target_os = "redox"))]
+use std::os::unix::prelude::*;
+#[cfg(any(unix, target_os = "redox"))]
+use std::collections::{HashMap, hash_map};
+#[cfg(any(unix, target_os = "redox"))]
+use std::ffi::OsString;
 use std::io;
 use std::path::Path;
 use std::io::prelude::*;
 use std::fs;
 use std::borrow::Cow;
-use std::collections::{HashMap, hash_map};
-use std::ffi::OsString;
 
 use {EntryType, Header, other};
 use header::{bytes2path, HeaderMode, path2bytes};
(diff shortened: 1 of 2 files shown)


BrianOn99:
I have fixed the warings in windows build.

alexcrichton:
Thanks for the PR! It looks like there's a fair bit of code movement here, could the code stay where it was primarily, outside the generic functions? Additionally could some restructuring perhaps help the amount of `#[cfg]` traffic?

BrianOn99:
There is quite a lot of code movement because a HashMap has to be accessed to find a hard link.  I moved those functions to methods so that the HashMap need not be passed around.  That makes the argument list shorter.  If that is not reasonable I can move them back.

As for the `#[cfg]`, the difficulty is I cannot find an open source implementation of tar on windows supporting hardlink.  I am not experienced on windows either.  Any suggestion will be appreciated!

alexcrichton:
Hm ok I definitely want to keep the implementations here outside of the generic functions, that may require some refactoring though to avoid passing too many parameters everywhere.

For Windows I do believe there's the basic equivalent of inodes. The [`same-file` crate](https://docs.rs/same-file/1.0.0/same_file/) may be useful here as well

BrianOn99:
@alexcrichton Thanks for the response.  I will move the implementations away, perhaps later this week.

As for windows, to my understanding using `same-file` may not be a good approach.  In unix we only record and find hard link when hard link count > 1, which save lots of memory and time.  And accessing HashMap is quite fast.  But to use `same-file` we have to record every file, and call `is_same_file` for every (recorded file, currentfile) pair, so it is O(n^2) for n = number of files.  Again I have no experience in windows API so I cannot think out a better approach.  Any expert here?

BrianOn99:
The requirements are:
1. `append_fs`, `append_path`, etc, have to be implemented outside `impl Buider`.
2. reduce number of `cfg`.
3. (my own requirement) not perform poorly when there are many files.

And in current situation I cannot find a way to meet all of them simultaneously.  To achieve [2] we can consider using `same-file` crate, but then [3] cannot be achieved, as mentioned above.  In the current state, trying to achieve [1] will violate [2] because I need to make version in of `append_fs` which takes a HashMap, and windows version does not take `HashMap`.

So I am closing this PR.  If anyone is interested feel free to use my commits.

PR #136, opened 2017-11-22
Set correct mtime when creating header

nossralf:
This makes tar-rs produce default headers that are standards-compliant.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -145,6 +145,7 @@ impl Header {
             gnu.magic = *b"ustar ";
             gnu.version = *b" \0";
         }
+        header.set_mtime(0);
         header
     }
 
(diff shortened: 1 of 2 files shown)


nossralf:
I've verified that cargo, when using tar-rs with this patch, will produce crate files that can be unpacked by 7-zip and thereby fixes rust-lang/cargo/#4712.

alexcrichton:
Awesome, thanks! Could you be sure to include a unit test here as well for this behavior?

nossralf:
I've added a test and realized I had forgotten to handle the old-style headers as well, so the PR now handles that correctly as well.

alexcrichton:
Thanks!

PR #143, opened 2017-12-29
Add a function to create headers from byte arrays without any transmute

est31:


est31:
After merging of this PR, I'd love to have a (minor) release.

est31:
It's not really important though as I can use a transmute for now.

alexcrichton:
This is a pretty large argument to be passing around, perhaps this could be switched to take a slice? We could also have it take a variable-length slice, assert it's 512 bytes, and then return a reference.

est31:
@alexcrichton is it better now? I'll need to test this version of the function first whether it actually works...

alexcrichton:
Yeah that looks fine by me, although could this take `&[u8]` instead of `&[u8; 512]`? I imagine the latter is very difficult to construct...

est31:
> I imagine the latter is very difficult to construct...

[Indeed](https://github.com/rust-lang/rfcs/issues/1772)... My code can manage as it serializes from/to the array directly, but I'll change it to `&[u8]`. As slices also need to store the length, I'll have to convert the slice to a pointer.

est31:
Please don't merge this yet, I haven't verified whether it does what I want it to. With unsafe code you never know...

est31:
Hmm right there seems to be an issue here where we invoke UB... Let me see whether I can fix this.

est31:
Yup, it is fixed now. r? @alexcrichton

alexcrichton:
Thanks!

PR #153, opened 2018-05-10
Initial attempt at async header writing

jjl:
Probably not ready for merging yet, but what do you think so far?

I'm still pretty new to tokio, so not quite sure if this is a good approach.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ contents are never required to be entirely resident in memory all at once.
 
 [dependencies]
 filetime = "0.2"
-
+tokio = { version = "0.1.6", optional = true }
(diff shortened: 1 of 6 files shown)


alexcrichton:
Thanks for the PR!

In looking at this though I think that we may want to take a slightly different route here. Or rather, I think we'll want an additional route as the low-level implementation. Currently this adds future combinators to write out the archive, but I think we'll want to also add low-level function to the main types for management with async APIs and such. 

That way the support isn't forced into using combinators when not necessary, and the underlying implementation should be more than ready for async I/O!

alexcrichton:
I'm gonna close this as it has gone a bit stale, but it can of course be resubmitted!

PR #158, opened 2018-07-23
Handle numeric extension for size

losynix:
Fix encoding of the size field when packing files larger than 8GB using tar's numeric extension.
Unpacking is already supported (#131).

Cf. the Numeric Extensions section : https://www.freebsd.org/cgi/man.cgi?query=tar&sektion=5&manpath=FreeBSD+8-current

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -307,6 +307,7 @@ impl Header {
         // stored in binary and the high bit of the first byte is set to 1
         // according to tar's Numeric Extensions.
         if size >= 8589934592 {
+            self.as_old_mut().size = [0; 12];
             self.as_old_mut().size[0] = 0x80;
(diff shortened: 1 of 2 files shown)


losynix:
I think it would be misleading to handle this special case in the **octal_{info, from}** functions because we're not really reading/writing octal data. Moreover the GNU tar format doesn't support this extension for all octal encoded fields so we should only check for it if the field is concerned.
For clarity and consistency I'd rather move the code of #131 from **octal_from()** to **entry_size()**, what do you think ?

alexcrichton:
Ah ok that sounds good to me! Want to throw that into this PR as well?

losynix:
yeap, I also added wrapper functions to handle the extension the same way with uid, gid and time fields.

alexcrichton:
Thanks!

PR #163, opened 2018-09-28
Always write big-endian in numeric_extended_into

### LINKED ISSUE #161: big-endian regression in 0.4.17
On Fedora Rawhide, every arch passed tests, except s390x failed on one:
https://koji.fedoraproject.org/koji/taskinfo?taskID=29929657
```
thread 'header::extended_numeric_format' panicked at 'assertion failed: `(left == right)`
  left: `[128, 0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0]`,
 right: `[128, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 1]`', tests/header/mod.rs:209:5
```

We no longer have other big-endian arches on rawhide, but I tried it directly on a Fedora 28 ppc64 host, and it fails the same way.  So it seems to be a general big-endian issue, not just s390x.

cuviper:
Before, this was transforming the number `to_be()`, then pulling out
least-significant bytes at shr 0, 8, 16, etc.  This has the effect of
writing reversed bytes on a little-endian target.  But since `to_be()`
does nothing on actual big-endian machines, they end up incorrectly
writing the true least-significant bytes first.

Instead, we can just reverse the order of the shift indexes, and then
the platform endianness is irrelevant.

Fixes #161.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -1422,7 +1422,7 @@ fn numeric_extended_into(dst: &mut [u8], src: u64) {
     let len: usize = dst.len();
     for (slot, val) in dst.iter_mut().zip(
         repeat(0).take(len - 8) // to zero init extra bytes
-            .chain((0..8).map(|x| ((src.to_be() >> (8 * x)) & 0xff) as u8)),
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #166, opened 2018-10-05
Strict entries for out of order processing

### LINKED ISSUE #160: Strict Entries for out-of-order (parallel) processing
Hi there.
I'm in the process of performance improvement for rustup, and have settled on extracting tars in parallel, having a POC showing its effectiveness on Windows.

Currently, the `tar` package only supports sequential unpacking of files to disk (`unpack` being the primitive operation as it also writes metadata).

To sum up the currently relevant motivations of the `tar` package, as I understand them:
1. Interface: Support any `Read` trait implementation, not assuming it's `Seek`.
2. Lazyness: Don't read data from the `Read`er unless the user asked to do so, instead call skip.

Both decisions are debatable, but are certainly very decent in the design space.

Now, in order to implement parallel extraction, I would need to process `Entry`ies out of order.

Solution space:
1. Make the interface require `Seek` - each entry can be independent, and still lazy, but this may potentially break clients.
2. Always be strict - with some performance hit only for complex applications that do filtering on the tar. Doesn't break the current api.
   - This may also simplify the API and avoid some mistakes people make by not reading the docs for `unpack` about out-of-order processing.
3. Add a `force` function on `Entry` - which returns a `StrictEntry`, being an independent structure, which reads & contains the data strictly. This would enlarge the user-facing API though - increasing implementation & api complexity. Doesn't break the current api.

I like both approach #2 and #3, each with their own trade-offs, and choosing which way to go depends on the crate's direction.

What are your thoughts?
1. What's your preferred solution to the problem?
2. Do you give a green light to start working on the suggested solution.

Thanks for the great crate.

NightRa:
Hi there.
This pull request exposes 2 new surface-level functions:
1. `Entry::force` - forces an Entry - reading it's contents into memory, allowing it's processing (unpack) in an out-of-order fashion.
2. `Archive::strict_entries` - provides an iterator of forced/strict entries.

This PR also refactors the internal code a bit to allow for this change.
- There's a change that separates the functions on `Entry` to those which require `Read` and those that don't, which clutters the diff a little bit.

Fixes #160

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rust:
   - stable
   - beta
   - nightly
-  - 1.21.0
+  - 1.26.0
 env:
   global:
     - RUSTFLAGS='-C link-dead-code'
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -2,7 +2,6 @@ use std::cell::{Cell, RefCell};
 use std::cmp;
 use std::io;
 use std::io::prelude::*;
-use std::marker;
 use std::path::Path;
 
 use entry::{EntryFields, EntryIo};
@@ -25,12 +24,7 @@ pub struct ArchiveInner<R: ?Sized> {
 }
 
 /// An iterator over the entries of an archive.
-pub struct Entries<'a, R: 'a + Read> {
-    fields: EntriesFields<'a>,
-    _ignored: marker::PhantomData<&'a Archive<R>>,
-}
-
-struct EntriesFields<'a> {
+pub struct Entries<'a> {
     archive: &'a Archive<Read + 'a>,
     next: u64,
     done: bool,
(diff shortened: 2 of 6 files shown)


NightRa:
Will fix the CI & the docs now.

NightRa:
I see that the projects uses rust `1.21` in the CI.
I used `impl Trait` which is stably supported from `1.26` forwards though.

What would you recomment to replace it with in `strict_entries`?
```rust
pub fn strict_entries<'a>(&'a mut self) -> io::Result< impl Iterator<Item=io::Result<Entry<Cursor<Vec<u8>>>>> + 'a >
```

alexcrichton:
Thanks for this! I'm feeling though that some of this is a bit too heavyweight for this library. For example `strict_entries` seems like it's largely the composition of `entries` and `force`, so I might not include that. For something like `force` I think I'd prefer a method which doesn't necessarily hardwire this to `Vec<u8>`. I could imagine, for example, that using platform-specific APIs an underlying `File` could be used for parallel extraction rather than a `Vec<u8>`.

To that end would it be possible to hook up the data via some other method for out of line parallel processing? For example if an `Entry` gave a start/end bounds, within the original stream, would that be enough? We could then provide helper functions for "write this `Header` plus this content to disk" which would do all sorts of validation and whatnot.

alexcrichton:
I'm gonna close this as it's gone a bit stale, but it can of course be resubmitted!

PR #172, opened 2018-11-08
Support BSD-style tar directories

udoprog:
Some old tar archives (notably constructed using bsd tar) indicate directories as a regular file, but with a trailing slash (`b"/"`).

Example: http://mirrors.kernel.org/pub/linux/kernel/Historic/linux-0.01.tar.gz

This mimics what gnu tar does to deal with this:
http://git.savannah.gnu.org/cgit/tar.git/tree/src/extract.c#n1547

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -394,21 +394,27 @@ impl<'a> EntryFields<'a> {
         Ok(true)
     }
 
+    /// Unpack as destination directory `dst`.
+    fn unpack_dir(&mut self, dst: &Path) -> io::Result<()> {
+        // If the directory already exists just let it slide
+        let prev = fs::metadata(dst);
+        if prev.map(|m| m.is_dir()).unwrap_or(false) {
+            return Ok(());
+        }
+        return fs::create_dir(dst).map_err(|err| {
+            Error::new(
+                err.kind(),
+                format!("{} when creating dir {}", err, dst.display()),
+            )
+        });
+    }
+
     /// Returns access to the header of this entry in the archive.
     fn unpack(&mut self, target_base: Option<&Path>, dst: &Path) -> io::Result<()> {
         let kind = self.header.entry_type();
+
         if kind.is_dir() {
-            // If the directory already exists just let it slide
-            let prev = fs::metadata(&dst);
-            if prev.map(|m| m.is_dir()).unwrap_or(false) {
-                return Ok(());
-            }
-            return fs::create_dir(&dst).map_err(|err| {
-                Error::new(
-                    err.kind(),
-                    format!("{} when creating dir {}", err, dst.display()),
-                )
-            });
+            return self.unpack_dir(dst);
         } else if kind.is_hard_link() || kind.is_symlink() {
             let src = match self.link_name()? {
                 Some(name) => name,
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! I wonder though if the `EntryType` should remain the same and perhaps just the extraction code could be updated to handle this case?

udoprog:
@alexcrichton I'll give it a shot!

udoprog:
@alexcrichton This looks better, thanks for the suggestion!

alexcrichton:
Looks good to me! As a final comment, could the function `name_has_trailing_slash` avoid being exported and the condition just tested inline when unpacking an entry?

udoprog:
Done!

src/entry.rs
```diff
@@ -489,6 +495,13 @@ impl<'a> EntryFields<'a> {
             return Ok(());
         };
 
+        // Old BSD-tar compatibility.
+        //
+        // Names that have a trailing slash should be treated as a directory.
+        if truncate(&self.header.as_old().name).ends_with(b"/") {
```
alexcrichton: Could this perhaps use something like `name_bytes()` so the `truncate` and `as_old` and such is already done internally?

src/error.rs
```diff
@@ -29,7 +29,7 @@ impl error::Error for TarError {
 
 impl fmt::Display for TarError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        self.desc.fmt(f)
+        write!(f, "{}: {}", self.desc, self.io)
```
alexcrichton: Oh for this it's intentional that the I/O error is omitted because it's hooked up in the chain of errors so if we print it here it'll get printed out in two locations

udoprog: Oh, I missed that. I wasn't seeing the error at all since my (generic) error handler only ran `fmt::Display` over them.

I'll revert this, thanks!

src/entry.rs
udoprog: Done!

src/header.rs
```diff
@@ -337,6 +337,11 @@ impl Header {
         bytes2path(self.path_bytes())
     }
 
+    /// Access the name portion of this header as a truncated slice of bytes.
+    pub fn name_bytes(&self) -> &[u8] {
```
alexcrichton: Er sorry I think I misremembered, but does `path_bytes` below work for you?

udoprog: Maybe, I'm not sure this applies to ustar paths as well. How do you think that should be handled?

alexcrichton: Could the conditional logic to look for the trailing slash only apply to non-ustar headers maybe? (or only one particular kind where this is known to come up?)

udoprog: What about something like this?

```
pub fn old_name_bytes(&self) -> Option<&[u8]> {
    if self.is_ustar() {
        None
    } else {
        Some(truncate(&self.as_old().name))
    }
}
```

src/header.rs
```diff
@@ -181,7 +181,7 @@ impl Header {
         header
     }
 
-    fn is_ustar(&self) -> bool {
+    pub(crate) fn is_ustar(&self) -> bool {
```
alexcrichton: Instead of making this `pub(crate)`, I think the same can be achieved with `header.as_ustar().is_some()`

udoprog:
I think that's it!

alexcrichton:
👍

PR #175, opened 2018-11-29
Add two new methods to tar Builder: inner_ref and flush_inner.

nickkuk:
These methods are useful while working with tar.gz:

- `inner_ref` allows to get current size of gzipped archive;

- `flush_inner` ensures that tar entry was written (look at [test](https://github.com/alexcrichton/tar-rs/compare/master...nickkuk:master#diff-1af2312a884fedee3dfc165a03ecfdd6R165)).

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -60,6 +60,16 @@ impl<W: Write> Builder<W> {
         Ok(self.obj.take().unwrap())
     }
 
+    /// Gets shared reference to the underlying object.
+    pub fn inner_ref(&self) -> &W {
+        self.obj.as_ref().unwrap()
+    }
+
+    /// Flushes the underlying object.
+    pub fn flush_inner(&mut self) -> io::Result<()> {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Could this perhaps be an idiomatic `get_ref` and `get_mut`? That way we wouldn't need to expose `flush_inner` as it could be composed with `get_mut`

nickkuk:
Ok, I can rename `inner_ref` to `get_ref`, but it seems to me that already existent method
```rust
    fn inner(&mut self) -> &mut W {
        self.obj.as_mut().unwrap()
    }
```
was hidden by design to not allow user to break tar archive somehow, so I decided to make public only `flush_inner` as it looks safe.

alexcrichton:
Ah yeah that was the original intention, but since then a convention has been developed that access is allowed and it's just documented that corruption can happen if misused

nickkuk:
Ok, let's try it in another PR :smile:

PR #176, opened 2018-11-29
Add two public methods to tar Builder: get_ref and get_mut.

nickkuk:
@alexcrichton, what do you think about this?


alexcrichton:
Looks good to me, thanks!

Issue #177, opened 2018-12-13
Usage Help: Updating An Existing Archive


guspower:
`tar:Archive` version `0.4.20` states:
> This archive can have an entry added to it

I am trying to append a file to an existing tar. Is this possible? If so, is there an example or test case demonstrating this functionality?

alexcrichton:
Thanks for the report, I think this is a bug in the documentation! Nowadays `Builder` is used to construct/append archives, but `Archive` is a read-only variant.

guspower:
Ah ok gotcha. Is there a working example or test case that shows a file being appended being appended to an _existing_ tar file using `Builder`? I can only see a way of creating new tar files using `Builder::new`.

alexcrichton:
I don't think there's an example currently, but that'd be a great one to add! You'll need to recreate the whole archive by iterating with `Archive`, adding to `Builder`, and then adding a final entry for now

jakebird451:
> You'll need to recreate the whole archive by iterating with `Archive`, adding to `Builder`, and then adding a final entry for now

@alexcrichton Is this still the preferred method for adding files to an archive? Adding in a file into an archive that is gigabytes large by dumping the original archive's contents into ram and writing out the content, with most of the contents identical to the original archive, seems a bit wasteful to me. Then again, I do not know the structure of a tar archive to know if this is even possible without needing to overwrite the entire archive.

I would also imagine that modifying a file that previously exists - or even deleting a file from an existing archive - would be a difficult task to accomplish without dumping the entire contents of the original archive into ram via the Builder, using a filter to remove unwanted files/modified files. Or maybe there is a less memory intensive method to perform actions like this?

alexcrichton:
A tar archive is "just" files concatenated together so if your archive is as big as you say it is then you should be able to seek to the end and use `Builder` to append more files.

jakebird451:
That is very interesting. Then going off that knowledge, to delete a file from the archive, one would need to: (1) find out where the file starts and ends in the archive and then (2) trim the file accordingly to no longer include that file. And conversely, to modify an existing file in the tar archive, you would effectively perform the same action as delete, but: (1) shrink/expand the space needed at the region in the tar of the old file to the size that fits the change and (2) write in the change.
[...]

PR #178, opened 2018-12-15
Enhancement: Append Existing Archive or Entry

guspower:
As per issue #177 here's a sample implementation with supporting tests and docs.

Let me know what you think.


src/builder.rs
```diff
@@ -111,6 +111,73 @@ impl<W: Write> Builder<W> {
...
+    /// let mut input = Archive::new(File::open("/path/to/existing.tar").unwrap());
+    ///
+    /// let mut output_file = File::create("/path/to/target.tar").unwrap();
+    /// let mut output = Builder::new(output_file);
+    ///
+    /// output.append_archive(&mut input).unwrap();
+    /// output.finish().unwrap();
+    /// ```
+    pub fn append_archive<R: Read>(&mut self, archive: &mut Archive<R>) -> io::Result<()> {
+        let entries = archive.entries()?;
```
alexcrichton: I think this may want to use `entries_raw` perhaps? That way it'll handle entries-describing-other-entries I think

src/builder.rs
```diff
@@ -111,6 +111,73 @@ impl<W: Write> Builder<W> {
...
+    /// let mut entry = input.entries().unwrap().next().unwrap().unwrap();
+    ///
+    /// let mut output_file = File::create("/path/to/target.tar").unwrap();
+    /// let mut output = Builder::new(output_file);
+    ///
+    /// output.append_entry(&mut entry).unwrap();
+    /// output.finish().unwrap();
+    /// ```
+    pub fn append_entry<R: Read>(&mut self, entry: &mut Entry<R>) -> io::Result<()> {
+        let header = entry.header().clone();
```
alexcrichton: I'm a little worried that this loses information because it doesn't handle things like long path names and such?

guspower: OK - can you explain this one a bit more? I see that in https://github.com/alexcrichton/tar-rs/blob/master/src/header.rs#L888 clone uses `self.bytes` - should I be checking `is_ustar` and `is_gnu` instead and constructing the appropriate `Header` instance?

Also could you describe an example that I could use as a test to verify this works as expected?

Cheers! G

alexcrichton: Oh sorry yeah, I should clarify! The `Entry<T>` type has a number of inherent methods:

* `Entry::path`
* `Entry::link_name`
* `Entry::path_extensions`

With various extensions to the `tar` format one instance of `Entry<T>` may actually be multiple tar-level entries in the underlying file. For example some entries just say "this is a longer path name than you could otherwise store". That's what the `raw(true)` for the entries iterator above is doing, it's avoiding automatically coalescing entries together and simply appending them one-by-one.

Here though an `Entry` may represent multiple tar-level entries that need to be emitted, but I believe the implementation here is only emitting one. Perhaps at least `path` could be handled? I think if you call `append_path` it should largely do the right thing.

guspower: Right I think I understand. I see looking at the [UStar format](https://en.wikipedia.org/wiki/Tar_(computing)#UStar_format) that the `'g'` and `'x'` header type flags can provide all sorts of stuff:
> extended header with meta data for the next file in the archive

I had no idea tar files could be that exciting!  

I'll add in some test cases that will hopefully surface the shortcoming in the implementation and then fix it up.

alexcrichton: Yeah tar files ended up being full of surprises, I thought writing a parser library like this wouldn't be too bad and here we are!

guspower: Yeah no kidding! 

I managed to replicate the issue with path truncation you mentioned, and have a simple change that handles long path names (as per your suggestion):
```self.append(&mut header, entry)?;``` -> ```self.append_data(&mut header, path, entry)?;```

The only issue I have is with a test assertion on the contents of the generated entry header - specifically with the checksum. It seems the `Builder` generates a different format checksum from `GNU tar`:

Builder | GNU tar
------------ | -------------
`0007666 `|`007666  `

i.e. the `Builder` creates a 7 digit representation followed by a space, while `GNU tar` creates a 6 digit representation followed by a NUL and then a space.

Wikipedia says:
> It is stored as a six digit octal number with leading zeroes followed by a NUL and then a space. Various implementations do not adhere to this format. For better compatibility, ignore leading and trailing whitespace, and take the first six digits. 

Do you think I should update the checksum implementation to generate `6 digits w/ leading zeros + NUL + space` so that it matches the `GNU tar` behaviour?

Comment (guspower) [reply src/builder.rs]: More specifically, I'm referring to the use of `octal_into` for checksum values https://github.com/alexcrichton/tar-rs/blob/master/src/header.rs#L1392 - I think the last two array values should be `NUL` and `space` (`[0, 32]`).

File: src/builder.rs [hunk 3]
```diff
@@ -111,6 +111,72 @@ impl<W: Write> Builder<W> {
...
+    /// let mut output = Builder::new(output_file);
+    ///
+    /// output.append_archive(&mut input).unwrap();
+    /// output.finish().unwrap();
+    /// ```
+    pub fn append_archive<R: Read>(&mut self, archive: &mut Archive<R>) -> io::Result<()> {
+        let entries = archive.entries()?.raw(true);
+
+        for entry in entries {
+            self.append_entry(entry?)?;
```
alexcrichton: I think this here, because it's a raw iterator, may actually want to use `append` directly for the reason that I think it's guaranteed to be correct

src/builder.rs
```diff
@@ -111,6 +111,72 @@ impl<W: Write> Builder<W> {
...
+        }
+
+        Ok(())
+    }
+
+    /// Appends an entry from an existing archive to this archive.
+    ///
+    /// This function will append an existing `Entry` to the builder target.
```
alexcrichton: Could some caveats be mentioned here about things like PAX metadata and I think long symlink names not being preserved just yet?

src/header.rs
```diff
@@ -677,7 +677,17 @@ impl Header {
     /// this header.
     pub fn set_cksum(&mut self) {
         let cksum = self.calculate_cksum();
-        octal_into(&mut self.as_old_mut().cksum, cksum);
+
+        let dst = &mut self.as_old_mut().cksum;
+        let o = format!("{:o}", cksum);
```
alexcrichton: This has historically been really finnicky, but this change seems fine to me! Perhaps though the implementation could still use `octal_into`, but the slice passed in could be two elements shorter which are manually initialized here? (avoids the need to duplicate some of the implementation)

guspower: yep I thought about pulling it all into `octal_into` but I thought it was a bit cryptic so I kept it separate (so you could more clearly see what was going on) - I'll update `octal_into` as you recommend.

src/builder.rs
guspower: I'm struggling to replicate GNU `@LongLink` or PAX metadata loss in a test. I've added some additional test resources - `pax3.tar` and `gnu-longname.tar` - to try and push these edge cases (see [append_entry_from_existing_archive](https://github.com/alexcrichton/tar-rs/blob/0fe050da0a6702373b06d8192393ee2393f91187/tests/all.rs#L403)). 

Any ideas about how I can enhance these tests to catch the unsupported cases?

src/builder.rs
guspower: Yep good point

.gitignore
```diff
@@ -1,2 +1,7 @@
 /target
 /Cargo.lock
+
+#Intellij Rust IDE
+.idea
+*.iml
+*.ipr
```
dvc94ch, line 7: don't think these should be here. here is some more information on how to configure git https://gist.github.com/subfuzion/db7f57fff2fb6998a16c locally to ignore those files.

dvc94ch (approved):
Great, this looks like a very useful feature.

guspower:
@dvc94ch If this PR is still considered useful I'm happy to bring it up-to-date.

dvc94ch:
That would be great. Should be fairly easy, using this branch for now: https://github.com/dvc94ch/tar-rs/tree/append-entry.

PR #181, opened 2019-02-01
Allow mtime field to be all nulls

FauxFaux:
`cargo` manages to generate `tar`s which have all nulls (i.e. the empty string) in the `mtime` field. This changes the code to accept this case, and treat it as if it was a valid `0`.

Here's `tar-rs` trying to read the metadata:

```
% cargo -q package --allow-dirty && \
   zcat target/package/*.crate | cargo -q run --example raw_list | \
   tail -n 6

path: tar-0.4.20/Cargo.toml
size: 1496
entry size: 1496
link name: None
file type: 0x30
mode: 0o644

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
  Custom { kind: Other, error: StringError(
    "numeric field was not a number:  when getting uid for tar-0.4.20/Cargo.toml") }', 
```

GNU tar is fine with this, it treats it as if the field contained `0`:

```
% zcat target/package/*.crate | tar vt | fgrep Cargo.toml
-rw-r--r-- 1000/1000       916 2019-01-29 21:53 tar-0.4.20/Cargo.toml.orig
-rw-r--r-- 0/0            1496 1970-01-01 01:00 tar-0.4.20/Cargo.toml
```

Here's the broken file in the tests. `mtime` is after the `length` (`02413`), and before the `checksum` (`0007360`).
```
00000e60: 0000 0000 3030 3030 3634 3400 0000 0000  ....0000644.....
00000e70: 0000 0000 0000 0000 0000 0000 3030 3030  ............0000
00000e80: 3030 3032 3431 3300 0000 0000 0000 0000  0002413.........
00000e90: 0000 0000 3030 3037 3336 3020 3000 0000  ....0007360 0...
```

Now.. the amusing part: `cargo` is using `tar-rs` to produce these archives.

It (`cargo 1.33.0-nightly (907c0febe 2019-01-20)`) is using modern enough `tar-rs` to pick up b8d44a3930e58c3c2ad9b121087f454a2d08830d: https://github.com/rust-lang/cargo/blame/d75d1fbe6b43e59d0a04e73659e411855c2633fc/Cargo.toml#L56

The usage looks correct enough to me:
https://github.com/rust-lang/cargo/blob/c9fa7db4d40c8fbf8b9f14c6a54f949cf4861ffd/src/cargo/ops/cargo_package.rs#L315

The `make_null_date` test does not have the issue. I am lost. This is a second attempt at #180, in which I became very confused.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -485,12 +485,14 @@ impl Header {
 
     /// Returns the last modification time in Unix time format
     pub fn mtime(&self) -> io::Result<u64> {
-        num_field_wrapper_from(&self.as_old().mtime).map_err(|err| {
-            io::Error::new(
+        match num_field_wrapper_from(&self.as_old().mtime) {
+            Ok(time) => Ok(time),
+            Err(ref err) if io::ErrorKind::NotFound == err.kind() => Ok(0),
+            Err(err) => Err(io::Error::new(
                 err.kind(),
                 format!("{} when getting mtime for {}", err, self.path_lossy()),
-            )
-        })
+            )),
+        }
     }
 
     /// Encodes the `mtime` provided into this header.
@@ -1372,7 +1374,7 @@ impl Default for GnuExtSparseHeader {
 fn octal_from(slice: &[u8]) -> io::Result<u64> {
     let trun = truncate(slice);
     if trun.is_empty() {
-        return Ok(0);
+        return Err(io::ErrorKind::NotFound.into());
     }
     let num = match str::from_utf8(trun) {
         Ok(n) => n,


alexcrichton:
Thanks for this! I'm somewhat confused though about how this is coming up, I haven't seen any reports in the wild about Cargo failing to extract files. Do you know how this could be triggered with this repository today?

FauxFaux:
I haven't worked it out, no. `cargo` *generates* invalid files, but can *extract* them fine, perhaps it ignores the metadata?

I'm interested in using `tar-rs` to unpack arbitrary files, some of which happens to have been made by `cargo`, and hence hit this issue and this fix.

alexcrichton:
Oh I think I see, Cargo doesn't look at the uid/gid field at all so if those are the only invalid fields then Cargo won't error on extraction, only when trying to exhaustively print everything.

In any case that all sounds good to me! I think there's some test failures on CI though?

FauxFaux:
I'd convinced myself that those tests were broken before my patch, and they don't seem to make any sense. I'll try a different machine.

FauxFaux:
I went with returning a different `io::ErrorKind`, and handling that in `mtime()`?

FauxFaux:
Maybe we shouldn't deal with it in `mtime`, and force clients to deal with it? I don't know what I'm doing, or what people would expect.

alexcrichton:
Hm ok, thanks for taking a look! I think that's probably related to this, right?

https://github.com/alexcrichton/tar-rs/blob/c7b31b28d42631a39af6b3f3f05891abc104fdd3/src/entry.rs#L563-L575

I wonder if we should just check for zero there and skip as well as if it's an error? Do you know what `tar` does for a `mode` field that is empty?

PR #183, opened 2019-02-16
Untar directory with archive perms as default

### LINKED ISSUE #134: Directory permissions
Seems like, that upon untar() the library is creating directories with the default rwxr-xr-x mode even if the tar itself contains different permissions.

Starting from here:
https://github.com/alexcrichton/tar-rs/blob/d7977e748aacadb8431293087eee4bdc99e225ed/src/entry.rs#L384

1. The code finds a directory entry
2. If the dir exists does nothing
3. Otherwise it creates it with the default permissions

A 4. step could be to set_mode the directory with the permissions of the tar entry. Or do I miss something?

joshleeb:
This PR modifies the default behavior for extracting a directory. Previously the extracted dir would have the system default perms. Now, the directory will have the same permissions as the tar file.

Resolves #134

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -401,12 +401,12 @@ impl<'a> EntryFields<'a> {
         if prev.map(|m| m.is_dir()).unwrap_or(false) {
             return Ok(());
         }
-        return fs::create_dir(dst).map_err(|err| {
+        fs::create_dir(dst).map_err(|err| {
             Error::new(
                 err.kind(),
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Could a test be added for this?

joshleeb:
> Could a test be added for this?

@alexcrichton test added 😄

alexcrichton:
Thanks!

PR #184, opened 2019-02-26
builder: handle long link names

losynix:
As already implemented for long path names, handle packing long link with the
GNU extension (entry type 'K') in append_fs function.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -157,7 +157,7 @@ impl<W: Write> Builder<W> {
         path: P,
         data: R,
     ) -> io::Result<()> {
-        prepare_header(self.get_mut(), header, path.as_ref())?;
+        prepare_header_path(self.get_mut(), header, path.as_ref())?;
         header.set_cksum();
         self.append(&header, data)
     }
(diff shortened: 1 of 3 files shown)


src/builder.rs
```diff
@@ -362,7 +362,20 @@ fn append_dir(dst: &mut Write, path: &Path, src_path: &Path, mode: HeaderMode) -
...
 
-fn prepare_header(dst: &mut Write, header: &mut Header, path: &Path) -> io::Result<()> {
+fn prepare_header(size: u64, entry_type: u8) -> Header {
+    let mut header = Header::new_gnu();
+    header.as_gnu_mut().unwrap().name[..13].clone_from_slice(b"././@LongLink");
+    header.set_mode(0o644);
+    header.set_uid(0);
+    header.set_gid(0);
+    header.set_mtime(0);
+    header.set_size(size + 1);
```
alexcrichton: I think this is just the same as befeore, but I feel like a comment was missing from before as well. If you can recall why the `+1` happens here, could a comment be added indicating as to why?

src/builder.rs
```diff
@@ -362,7 +362,20 @@ fn append_dir(dst: &mut Write, path: &Path, src_path: &Path, mode: HeaderMode) -
     append_fs(dst, path, &stat, &mut io::empty(), mode, None)
 }
 
-fn prepare_header(dst: &mut Write, header: &mut Header, path: &Path) -> io::Result<()> {
+fn prepare_header(size: u64, entry_type: u8) -> Header {
+    let mut header = Header::new_gnu();
+    header.as_gnu_mut().unwrap().name[..13].clone_from_slice(b"././@LongLink");
```
alexcrichton: While you're here (I realize this was just copied from before), mind cleaning this up a bit? Ideally the string and the 13 length it has would be deduplicated slightly to just store the string to a local variable and read its lenght

src/builder.rs
```diff
@@ -391,6 +396,21 @@ fn prepare_header(dst: &mut Write, header: &mut Header, path: &Path) -> io::Resu
...
 
+fn prepare_header_link(dst: &mut Write, header: &mut Header, link_name: &Path) -> io::Result<()> {
+    // Same as previous function but for linkname
+    if let Err(e) = header.set_link_name(&link_name) {
+        let data = path2bytes(&link_name)?;
+        if data.len() < header.as_old().linkname.len() {
+            return Err(e);
+        }
+        let mut header2 = prepare_header(data.len() as u64, b'K');
+        let mut data2 = data.chain(io::repeat(0).take(0));
```
alexcrichton: I think this was copied from before (sorry for lots of comments like this) but I actually have no idea why `io::repeat(0).take(0)` is above (and copied down here). It seems like the `.chain` should be safe to delete, right?

src/builder.rs
```diff
@@ -391,6 +396,21 @@ fn prepare_header(dst: &mut Write, header: &mut Header, path: &Path) -> io::Resu
     Ok(())
 }
 
+fn prepare_header_link(dst: &mut Write, header: &mut Header, link_name: &Path) -> io::Result<()> {
+    // Same as previous function but for linkname
+    if let Err(e) = header.set_link_name(&link_name) {
+        let data = path2bytes(&link_name)?;
+        if data.len() < header.as_old().linkname.len() {
```
alexcrichton: Man really all I have is comments on the old code... So along those lines I actually have no idea what this is checking, but it seems like it's trying to check for errors not related to length, but for that I think it might be more appropriate to inspect the `e` error we got instead of checking lengths afterwards

src/builder.rs
losynix: I don't know either, it was introduced by this commit https://github.com/alexcrichton/tar-rs/commit/bbbe5b129099e214fb74aba4957d74fd97c6846c : "Apparently GNU does this, so we should too!". That's [true](https://git.savannah.gnu.org/cgit/tar.git/tree/src/create.c#n541), but I don't know why exactly, maybe to have a null-terminated name ?
I added a comment.

src/builder.rs
losynix: done !

src/builder.rs
losynix: Again, not sure why this was here, `.take(1)` would have made sense for the null-terminated name I mentioned in the other comment. This isn't a problem when name is shorter than 512 since `append()` will pad with 0 but I don't know if this last null byte is mandatory. 
Thing is `append()` wants something that implements `Read` so we can't just delete the `.chain` and call the function with `data` directly, would a `Cursor` be better ?

src/builder.rs
losynix: It's verifying that `set_link_name` failed because of the length being too long, if not, return `e`. We get `ErrorKind::Other` (from `copy_into` in src/header.rs)  if the name is too long, which isn't specific enough to differentiate from other errors unfortunately.

src/builder.rs
alexcrichton: Hm let's actually switch this to `io::repeat(0).take(1)`, as I agree that was probably the original intent and it's been "accidentally working" so far otherwise.

src/builder.rs
alexcrichton: Ok yeah that's what I figured. Also agreed there's not a great way to distinguish errors for now, but mind leaving a comment indicating what this pattern is doing?

src/builder.rs
losynix: alright

src/builder.rs
losynix: Sure

losynix:
One more thing, I need a public `append_` like function that supports symlinks and setting a different filename in the archive. `append_path()` can do the first requirement but not the second, `append_file()` can do the second but not the first. `append_dir_all()` can do both but I don't always need to add every files in a directory.
How do you think I should do this ?
- add a `name` argument to `append_path()` (unlikely cause it would break the current API) ?
- make `append_file()` "link aware" ?
- add a new `append_path_with_name()` (or something like that) public function ?

alexcrichton:
I'd personally prefer to avoid breakage of the current API, but the current trend is to just add suites of helper functions to the builder which all fall back to the raw `append` function. Adding another helper seems fine by me!

PR #185, opened 2019-03-03
builder: append_path_with_name()

losynix:
Add a new public function `append_path_with_name()`, similar to `append_path()` but allows adding files under a different name.


alexcrichton:
Nice!

Issue #147, opened 2018-01-30
append_dir_all appears to be working, but is returning an error


nellshamrell:
Hello!

I'm using append_dir_all to recursively add a directory and all subdirectories to a tar archive list this:

```
    let tarball_name = "my_tarball.gz";

    let tarball = File::create(tarball_name).unwrap();
    let mut tar_builder = Builder::new(tarball);

    let hab_pkgs_path = temp_dir_path.clone().join("rootfs/hab");

    tar_builder.append_dir_all("hab", hab_pkgs_path);
```

The odd thing is this code DOES work - it recursively adds the directory to the tar archive - but this line

```
tar_builder.append_dir_all("hab", hab_pkgs_path);
```

returns this error "thread 'main' panicked at 'could not export.: "Is a directory (os error 21)"', /checkout/src/libcore/result.rs:906:4"

I have been trying different things, googling left and right, running backtraces, and I have been unable to find a solution to this error.  Do you know of anything in this code that might cause that?

alexcrichton:
Oh dear if that panics then it's definitely a bug! Can you run with `RUST_BACKTRACE` and gist that?

nellshamrell:
Absolutely!  Here is the output from running with RUST_BACKTRACE

(Note - I am running the code above from a crate called hab_pkg_export_tar and a function called by another function called "export")

https://gist.github.com/nellshamrell/e4d9945c2a38f80cf13c345f39c93d23

alexcrichton:
Oh ok, so the panic it seems is coming from the application code? Do you perhaps have a test case I could reproduce this error with?

boastful-squirrel:
I think #48 solves this issue

alexcrichton:
Thanks!

PR #187, opened 2019-03-09
Handles symlinks pointing to directories

### LINKED ISSUE #174: Crashes when directory contains symlink to a directory. (When `append_dir_all`.)
Hi @alexcrichton! Respect your hard works on Rust!

I've found a bug for the tar crate. This is a small project to re-create the error.

https://github.com/tbrand/tar-error

The error occurs when source directory contains a symlink **to a directory**.
It might be related [this issue](https://github.com/alexcrichton/tar-rs/issues/147).
I hope this will be fixed soon. :smile:

Thanks!

boastful-squirrel:
This is an attempt to fix #174 and #147 

Note that I'm quite a beginner so I'm open to any criticism on the code quality or anything else. I tested my code on a directory containing a symlink to a directory, containing a symlink to a directory. I'm pretty confident my code can be optimised so it matches the first condition, though I can't seem to a find a predicate that would match... (`if is_dir || ????`).

If you need me to write unit tests associated with this features, I can do it

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -489,7 +489,8 @@ fn append_dir_all(
     let mut stack = vec![(src_path.to_path_buf(), true, false)];
     while let Some((src, is_dir, is_symlink)) = stack.pop() {
         let dest = path.join(src.strip_prefix(&src_path).unwrap());
-        if is_dir || is_symlink && follow && src.is_dir() {
+        // In case of a symlink pointing to a directory, is_dir is false, but src.is_dir() will return true
+        if is_dir || (is_symlink && follow && src.is_dir()) {
             for entry in fs::read_dir(&src)? {
                 let entry = entry?;
                 let file_type = entry.file_type()?;
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 use std::fs;
 use std::io;
 use std::io::prelude::*;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use header::{bytes2path, path2bytes, HeaderMode};
 use {other, EntryType, Header};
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! I was thinking the same thing though that ideally the refactored `browse_dir` function would only be called from one location because it seems odd for a non-file to leak into the "this is a file" clause later on.

Additionally yeah tests for this would be greatly appreciated!

boastful-squirrel:
So I think this should do it... It passes the tests, and my use case. I'll try to write tests, I never did so I can't promise anything.

boastful-squirrel:
So I wrote a very basic test, as I said I never did it before so I'm open to any comments. This test fails on `master`, and succeeds on this branch.

src/builder.rs
```diff
@@ -489,7 +489,7 @@ fn append_dir_all(
     let mut stack = vec![(src_path.to_path_buf(), true, false)];
     while let Some((src, is_dir, is_symlink)) = stack.pop() {
         let dest = path.join(src.strip_prefix(&src_path).unwrap());
-        if is_dir {
+        if is_dir || is_symlink && follow && src.is_dir() {
```
alexcrichton: Could parentheses be used here around the added clause to explicitly show the precedence? Additionally could you add a comment as well saying that `is_dir` being false is about the symlink, and the call to `.is_dir()` reaches through the symlink?

alexcrichton:
Looks great! Just a few minor nits but otherwise looks good to go to me

boastful-squirrel:
Hey, here you go :)

alexcrichton:
Thanks!

boastful-squirrel:
Do you plan on releasing 0.4.22 with this fix by the way ?

alexcrichton:
Done!

Issue #191, opened 2019-04-18
Compilation to WebAssembly (wasm-bindgen) failing due to filetime dependency


mpizenberg:
Hi again ^^, I just added tar as a dependency to a minimalist wasm skeleton code and compilation failed. It appears that the failing happens when compiling the filetime crate.

```
$> wasm-pack build --target web

...

error[E0433]: failed to resolve: could not find `unix` in `os`
 --> /home/matthieu/.cargo/registry/src/github.com-1ecc6299db9ec823/filetime-0.2.4/src/unix/mod.rs:6:14
  |
6 | use std::os::unix::prelude::*;
  |              ^^^^ could not find `unix` in `os`
```

Is this something that can be bypassed? Should I report the issue directly to the filetime crate?

I set up a minimalist github repo that you can just clone and build to test things: https://github.com/mpizenberg/wasm-tar

alexcrichton:
Ah that actually looks like the `filetime` crate is the one not compiling, should be easy enough though I think to send a PR there to ensure it compiles but just returns errors!

mpizenberg:
I've found those two pages [1,2] of the rustwasm book, that helped me understand better what hinders wasm compatibility. I don't understand however what you mean by "ensure it compiles but just returns errors".

From what I gather though, system librairies and file interactions are things not working with wasm. And so when I compile filetime with `cargo build --target wasm32-unknown-unknown`, I get a lot of "unresolved imports", "cannot find value '...' in module libc", etc. Adding wasm support to the filetime crate doesn't seem obvious to me.

In any case, I think (let me know if I'm wrong) that filetime is only used for `unpack` in the tar-rs crate. And unpacking should not be useful for wasm target. Is there a way of not providing the unpacking functionalities for the wasm32-unknown-unkown target?

[1] https://rustwasm.github.io/docs/book/reference/which-crates-work-with-wasm.html
[2] https://rustwasm.github.io/docs/book/reference/add-wasm-support-to-crate.html

mpizenberg:
I think I got what you meant! I was looking for other crates with the same issues and found a wasm PR for clap: https://github.com/clap-rs/clap/pull/1190/commits/689949e57d390bb61bc69f3ed91f60a2105738d0

It contained things like:

```
#[cfg(all(feature = "debug", not(target_arch = "wasm32")))]
use std::os::unix::ffi::OsStrExt;

#[cfg(not(any(target_os = "windows", target_arch = "wasm32")))]
[...]

PR #193, opened 2019-04-21
Partial wasm implementation (not crashing at build anymore)

mpizenberg:
When combined with https://github.com/alexcrichton/filetime/pull/39 it enables usage of tar-rs in WebAssembly. Some functions have an implementation, others less critical are left with `unimplemented!()`.


mpizenberg:
A minimalist wasm example logging all tar entries paths is available here: https://github.com/mpizenberg/wasm-tar

alexcrichton:
Thanks! Like with the filetime PR though could this return errors instead of panicking where possible?

mpizenberg:
Ok I've made the modifications for functions returning `io::Result` here and in filetime. They now return an error instead of panicking.

mpizenberg:
fix #191

Issue #198, opened 2019-05-09
privilege escalation vectors


rbtcollins:
Right now unpacking has a number of correctness-and-performance defects; in particular:
- checking for the presence of a dir before making one 
   (this is incorrect because an existing dir may have a different mode than a newly made one would have), and a performance issue because its an unneeded syscall in the common case.
- removing existing files before creating new files (this is a performance issue because its unneeded in the common case, and unsafe because it doesn't solve the race condition - a racing attacker with write access to the same directory could create a symlink between the unlink and the create and still cause an indirection attack to gain escalated privileges )
- permissions and xattrs are being done on path rather than the filehandle of the created file; this permits an attacker to move a different file into place to receive the unpacked mode/attributes (so for instance an archive with a setuid root file that is safe would permit applying setuid root to a malicious file).

Separately, but similar in that they are using the fs interface incorrectly/inefficiently...
- mtime setting is done via path, not the filehandle of the written file

The performance impact of these additional file operations is trivial for archives with large files, as it is lost in the work-per-file, but for archives with many small files, it can become a significant part of the overheads.

alexcrichton:
Thanks for the report! Good point on the performance issues, but I don't consider there to be any security/privilege issues here with concurrency. The "threat model" I don't think reasonably encompasses concurrent writes to the filesystem while we're extracting something, but rather extracting something to a location that we're the only one writing at the time. I think it's fine to have improvements in the name of speed which reduce the amount of sycalls, but I would not consider them security fixes.

rbtcollins:
Thank you for merging my initial patch.

On the security side, declaring a formal threat model for tar-rs would help in analysing what situations tar-rs would be suitable for. Excluding concurrent operations on the file system is a great simplifier. But I suspect that many users of tar-rs would not be able to make that simplification. For rustup we could change things to permit us to make it (by changing permissions on ~/.rustup/tmp), but it would be false today as we don't ensure that that path is secured. If a lower privilege user somehow got a handle into a higher privileged users ~/.rustup/tmp/ without these races fixed, they could do a lot of damage. Anyhow - if tar-rs wants to declare that 'single operator in the extraction directory' is the security model, that is fine, but please put it in README or something :).

Because there have been many many TOC[T]TOU style exploits over the years - https://www.usenix.org/legacy/event/fast05/tech/full_papers/wei/wei.pdf is a very old paper on them - and their feasibility and effectiveness is well established.

[...]

PR #199, opened 2019-05-10
Some time of check/time of use bugfixes

rbtcollins:
As described in bug #198 there are attack vectors in tar-rs at the moment. This patch corrects the worst of them, but doesn't go the whole way - changes to Filetime would be needed (or just not using that API at all) and I figured a discussion about that would make sense before going further.


alexcrichton:
Thanks!

PR #201, opened 2019-05-11
Perform file mode setting on the open file

rbtcollins:
Testing with rustup on Windows this reduces extraction of rust-docs
in my testbed from 26/27s to 22/23s, for similar reasons that using the
set_preserve_mtime(false) call there took it from 31/32s to 26/27s,
file system syscall efficiency matters a great deal on Windows.

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -420,7 +420,7 @@ impl<'a> EntryFields<'a> {
             return self
                 .unpack_dir(dst)
                 .and_then(|_| self.header.mode())
-                .and_then(|mode| set_perms(dst, mode, self.preserve_permissions));
+                .and_then(|mode| set_perms(dst, None, mode, self.preserve_permissions));
         } else if kind.is_hard_link() || kind.is_symlink() {
             let src = match self.link_name()? {
                 Some(name) => name,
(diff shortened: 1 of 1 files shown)


src/entry.rs
```diff
@@ -556,7 +556,8 @@ impl<'a> EntryFields<'a> {
                     }
                 }
             }
-            Ok(())
+            f.flush()?;
```
alexcrichton: I would prefer that this were left out unless there's a comment indicating why it's necessary.

I don't believe this is necessary because the [`flush` implementation on Windows](https://github.com/rust-lang/rust/blob/a9ec99f4201ec33026a468ef1289f98a95b4d71a/src/libstd/sys/windows/fs.rs#L337) is an empty function.

src/entry.rs
```diff
@@ -597,24 +598,51 @@ impl<'a> EntryFields<'a> {
...
+            _preserve: bool,
+        ) -> io::Result<()> {
+            match f {
+                Some(f) => {
+                    let mut perm = f.metadata()?.permissions();
+                    perm.set_readonly(mode & 0o200 != 0o200);
+                    f.set_permissions(perm)
+                }
+                None => {
+                    let mut perm = r#try!(fs::metadata(dst)).permissions();
```
alexcrichton: FWIW if this is run across it's fine to replace all instances of `r#try!` with `?`

src/entry.rs
rbtcollins: It's also empty on unix; I'll move the comment here - it being a no-op today isn't a guarantee it will be a no-op tomorrow, and as file permissions are external state I would expect the type system rust has around files to save us (that would be a rather bigger change), for set_permissions to consume the File and return a new one...

alexcrichton: Er sorry, so to be clear, I wrote that part of the standard library. The `flush` here is not needed and is distracting from what this function is doing. The `flush` method is simply flushing buffered written data for buffered writers, nothing else. There's no magical buffers for permission bits or things like that, so this needs to be removed.

rbtcollins: Easy fix then :).

PR #202, opened 2019-05-12
WIP: Eliminate readonly syscalls for files on Windows

rbtcollins:
The Windows API permits files to be created with the desired attributes
rather than setting them via a later syscall.

This is already exposed in Rust via the std::os modules. I can add a use of winapi for the
constant if desired.

This is a follow-on to #201 - they textually conflict or I would have made this fully standalone.

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -525,17 +525,32 @@ impl<'a> EntryFields<'a> {
 
         // Ensure we write a new file rather than overwriting in-place which
         // is attackable; if an existing file is found unlink it.
-        fn open(dst: &Path) -> io::Result<std::fs::File> {
-            OpenOptions::new().write(true).create_new(true).open(dst)
+        let mode = self.header.mode().ok();
+        #[allow(unused_variables)]
+        fn open(dst: &Path, mode: Option<u32>) -> io::Result<std::fs::File> {
+            let mut options = OpenOptions::new();
+            #[cfg(windows)]
+            {
+                use std::os::windows::fs::OpenOptionsExt;
+                use winapi::um::winnt;
+
+                // On windows create the file with the desired attributes rather
+                // than setting them in a separate syscall.
+                if !user_can_write(mode) {
+                    // File is requested to be readonly
+                    options.custom_flags(winnt::FILE_ATTRIBUTE_READONLY);
+                }
+            }
+            options.write(true).create_new(true).open(dst)
         };
         let mut f = (|| -> io::Result<std::fs::File> {
-            let mut f = open(dst).or_else(|err| {
+            let mut f = open(dst, mode).or_else(|err| {
                 if err.kind() != ErrorKind::AlreadyExists {
                     Err(err)
                 } else {
                     match fs::remove_file(dst) {
-                        Ok(()) => open(dst),
-                        Err(ref e) if e.kind() == io::ErrorKind::NotFound => open(dst),
+                        Ok(()) => open(dst, mode),
+                        Err(ref e) if e.kind() == io::ErrorKind::NotFound => open(dst, mode),
                         Err(e) => Err(e),
                     }
                 }
@@ -578,7 +593,7 @@ impl<'a> EntryFields<'a> {
                 })?;
             }
         }
-        if let Ok(mode) = self.header.mode() {
+        if let Some(mode) = mode {
             set_perms(dst, Some(&mut f), mode, self.preserve_permissions).map_err(|e| {
                 TarError::new(
                     &format!(
@@ -621,17 +636,22 @@ impl<'a> EntryFields<'a> {
             _preserve: bool,
         ) -> io::Result<()> {
             match f {
-                Some(f) => {
-                    let mut perm = f.metadata()?.permissions();
-                    perm.set_readonly(mode & 0o200 != 0o200);
-                    f.set_permissions(perm)
+                Some(_f) => {
+                    // Set at file creation
+                    Ok(())
                 }
                 None => {
                     let mut perm = fs::metadata(dst)?.permissions();
-                    perm.set_readonly(mode & 0o200 != 0o200);
+                    perm.set_readonly(!user_can_write(Some(mode)));
                     fs::set_permissions(dst, perm)
                 }
             }
+        }
+        #[cfg(windows)]
+        fn user_can_write(mode: Option<u32>) -> bool {
+            mode.as_ref()
+                .map(|mode| mode & 0o200 == 0o200)
+                .unwrap_or(false)
         }
 
         #[cfg(target_arch = "wasm32")]
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -573,7 +573,7 @@ impl<'a> EntryFields<'a> {
         if self.preserve_mtime {
             if let Ok(mtime) = self.header.mtime() {
                 let mtime = FileTime::from_unix_time(mtime as i64, 0);
-                filetime::set_file_times(dst, mtime, mtime).map_err(|e| {
+                filetime::set_file_handle_times(&f, Some(mtime), Some(mtime)).map_err(|e| {
                     TarError::new(&format!("failed to set mtime for `{}`", dst.display()), e)
                 })?;
             }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ contents are never required to be entirely resident in memory all at once.
 """
 
 [dependencies]
-filetime = "0.2"
+filetime = "0.2.6"
 
 [dev-dependencies]
 tempdir = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,9 @@ tempdir = "0.3"
 xattr = { version = "0.2", optional = true }
 libc = "0.2"
 
+[target.'cfg(windows)'.dependencies]
+winapi = { version = "0.3", features = ["winnt"] }
+
 [target.'cfg(target_os = "redox")'.dependencies]
 redox_syscall = "0.1"
 
--- a/src/header.rs
+++ b/src/header.rs
@@ -781,6 +781,7 @@ impl Header {
 
     #[cfg(windows)]
     fn fill_platform_from(&mut self, meta: &fs::Metadata, mode: HeaderMode) {
+        use winapi::um::winnt::FILE_ATTRIBUTE_READONLY;
         // There's no concept of a file mode on windows, so do a best approximation here.
         match mode {
             HeaderMode::Complete => {
@@ -793,7 +794,6 @@ impl Header {
                 let mtime = (meta.last_write_time() / (1_000_000_000 / 100)) - 11644473600;
                 self.set_mtime(mtime);
                 let fs_mode = {
-                    const FILE_ATTRIBUTE_READONLY: u32 = 0x00000001;
                     let readonly = meta.file_attributes() & FILE_ATTRIBUTE_READONLY;
                     match (meta.is_dir(), readonly != 0) {
                         (true, false) => 0o755,
--- a/tests/all.rs
+++ b/tests/all.rs
@@ -9,6 +9,7 @@ use std::io::prelude::*;
 use std::io::{self, Cursor};
 use std::iter::repeat;
 use std::path::{Path, PathBuf};
+use std::time::{Duration, SystemTime};
 
 use self::tempdir::TempDir;
 use filetime::FileTime;
@@ -842,6 +843,48 @@ fn extract_sparse() {
     assert!(s[0x1000 + 6..0x2fa0].chars().all(|x| x == '\u{0}'));
     assert_eq!(&s[0x2fa0..0x2fa0 + 6], "world\n");
     assert!(s[0x2fa0 + 6..0x4000].chars().all(|x| x == '\u{0}'));
+}
+
+#[test]
+fn extracting_readonly_file_with_mtime() -> io::Result<()> {
+    let td = t!(TempDir::new("tar-rs"));
+
+    let mut tar = Vec::new();
+
+    {
+        let mut a = Builder::new(&mut tar);
+        let path = "file.txt";
+
+        let mut header = Header::new_gnu();
+        header.set_path(path)?;
+        {
+            let h = header.as_gnu_mut().unwrap();
+            for (a, b) in h.name.iter_mut().zip(path.as_bytes()) {
+                *a = *b;
+            }
+        }
+        header.set_size(1);
+        header.set_mtime(2);
+        header.set_mode(0o400);
+        header.set_cksum();
+        t!(a.append(&header, io::repeat(1).take(1)));
+    }
+
+    let mut ar = Archive::new(&tar[..]);
+    ar.unpack(td.path())?;
+
+    let path = td.path().join("file.txt");
+    let m = fs::metadata(&path)?;
+    assert_eq!(
+        Duration::from_secs(2),
+        m.modified()?
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap()
+    );
+    assert!(m.permissions().readonly());
+    assert_eq!(1, fs::read(&path)?[0]);
+
+    Ok(())
 }
 
 #[test]


src/entry.rs
```diff
@@ -525,17 +527,29 @@ impl<'a> EntryFields<'a> {
...
+        let mode = self.header.mode();
+        fn open(dst: &Path, _mode: &io::Result<u32>) -> io::Result<std::fs::File> {
+            let mut options = OpenOptions::new();
+            #[cfg(windows)]
+            {
+                // On windows create the file with the desired attributes rather
+                // than setting them in a separate syscall.
+                if readonly(_mode) {
+                    // File is requested to be readonly
+                    // 0x1 is FILE_ATTRIBUTE_READONLY in the Win32 API
```
alexcrichton: Could this constant be loaded from winapi?

src/entry.rs
```diff
@@ -525,17 +527,29 @@ impl<'a> EntryFields<'a> {
 
         // Ensure we write a new file rather than overwriting in-place which
         // is attackable; if an existing file is found unlink it.
-        fn open(dst: &Path) -> io::Result<std::fs::File> {
-            OpenOptions::new().write(true).create_new(true).open(dst)
+        let mode = self.header.mode();
```
alexcrichton: Could this use `.ok()` to discard the error since it's ignored?

src/entry.rs
```diff
@@ -622,18 +636,23 @@ impl<'a> EntryFields<'a> {
...
                 None => {
                     let mut perm = r#try!(fs::metadata(dst)).permissions();
-                    perm.set_readonly(mode & 0o200 != 0o200);
+                    perm.set_readonly(readonly(&Ok(mode)));
                     fs::set_permissions(dst, perm)
                 }
             }
         }
+        #[cfg(windows)]
+        fn readonly(mode: &io::Result<u32>) -> bool {
```
alexcrichton: Could this change to `fn(u32) -> bool` since that's basically what's being deduplicated here?

alexcrichton: also could the name perhaps be something like `is_readonly_for_win32` or something like that?

rbtcollins:
> 
> 
> Out of curiosity how many readonly files are in rust-lang/rust tarballs? I figured there wouldn't really be that many in general.

None. But windows really doesn't like multiple mutating opens of the same file, so to avoid causing errors for folk that do have readonly files I wanted to get out in front of this.

src/entry.rs
rbtcollins: Sure thing; didn't want to add a large dep without discussion.

src/entry.rs
rbtcollins: that would be more awkward I think - because we only ignore the error to decide that readonly is false, so we'd be pulling apart the readonly logic  - we'd have to immediately evaluate the readonliness; whereas by keeping the monad we can use mode for the unix code paths as well.

src/entry.rs
rbtcollins: I have renamed it to user_can_write - I was staring at it and this jumped out at me; let me know if you would still prefer is_readonly_for_win32 : the function isn't actually win32 specific, its just only used on win32 today.

I don't want to change the signature though, because this cascades into the secure open function and it gets messy: that can't be a closure because there is already a closure borrowing self; as a function passing in windows specific parameters like 'readonly' will make the code hard to maintain as more os-specific cases occur, so I'd rather pass in only the actual data that is being operated on - e.g. the mode, and as mode is a monad, not a value, keeping the monad makes a lot of sense to me.

rbtcollins:
I'm not sure why this new build failed; presumably something to do with winapi being included, though the code should be the same :/. I will investigate. #201 should still be mergable.

rbtcollins:
reproduced locally

alexcrichton:
Hm ok if this isn't tested by rustup can a test be added here for extraction of a readonly file to ensure that it works as expected on Windows?

src/entry.rs
```diff
@@ -5,9 +5,13 @@ use std::fs::OpenOptions;
 use std::io::prelude::*;
 use std::io::{self, Error, ErrorKind, SeekFrom};
 use std::marker;
+#[cfg(windows)]
+use std::os::windows::fs::OpenOptionsExt;
 use std::path::{Component, Path, PathBuf};
 
 use filetime::{self, FileTime};
+#[cfg(windows)]
```
alexcrichton: Instead of `#[cfg]` imports, can these imports be placed in the block that's already `#[cfg]` windows below?

src/entry.rs
alexcrichton: Er my point is that passing around `&io::Result<u32>` is just pretty weird, so it would be more idiomatic to pass around `Option<u32>`

src/entry.rs
```diff
@@ -525,17 +529,28 @@ impl<'a> EntryFields<'a> {
 
         // Ensure we write a new file rather than overwriting in-place which
         // is attackable; if an existing file is found unlink it.
-        fn open(dst: &Path) -> io::Result<std::fs::File> {
-            OpenOptions::new().write(true).create_new(true).open(dst)
+        let mode = self.header.mode();
+        fn open(dst: &Path, _mode: &io::Result<u32>) -> io::Result<std::fs::File> {
```
alexcrichton: This should not be named `_mode`, but rather it should be called `mode`. I understand it's not used on all platforms, that warning should be fixed via method that does not involve changing the variable name.

src/entry.rs
alexcrichton: Let's please change this to `u32 -> bool`, unnecessarily having to wrap this in `&Ok(...)` in one callsite is pretty wonky.

src/entry.rs
rbtcollins: Oh, I get you now, thank you.

src/entry.rs
rbtcollins: allow(unused_variables) done instead. I had used _mode as it was narrower ,whereas allowing unused variables will permit other unused variables in future, at least as I understand it.

src/entry.rs
rbtcollins: Done

rbtcollins:
> 
> 
> Hm ok if this isn't tested by rustup can a test be added here for extraction of a readonly file to ensure that it works as expected on Windows?

There is a test catching it locally already, but not dedicated; once I understand why the test (which seems unrelated) is catching it, I will assess whether a dedicated one is needed or not and add if so.

The evil test which isn't meant to extract any files *at all* is erroring, which means it is failing in file operations... that aren' t meant to be taking place at all. So thats got me worrying about both the evil-detection logic and the test : I'll look into this and come back to this patch once I've got a proper understanding of the situation. Similarly name_with_slash_doesnt_fool_long_link_and_bsd_compat is behaving the same.

src/entry.rs
rbtcollins: Converted to Option<u32>

src/entry.rs
rbtcollins: I'll have a go at this, I started there and worked in to end up with the shape I got, but I'll see if I can make it not too ugly.

src/entry.rs
```diff
@@ -525,17 +525,32 @@ impl<'a> EntryFields<'a> {
 
         // Ensure we write a new file rather than overwriting in-place which
         // is attackable; if an existing file is found unlink it.
-        fn open(dst: &Path) -> io::Result<std::fs::File> {
-            OpenOptions::new().write(true).create_new(true).open(dst)
+        let mode = self.header.mode().ok();
+        #[allow(unused_variables)]
```
alexcrichton: Actually now that I think about it, Unix also has `OpenOptionsExt::mode` which I think could be used here?

rbtcollins: IIRC that interacts poorly with umask (compared to the semantics expected from e.g. bsd-tar/gnu-tar etc). I'm happy to look into that as a separate patch however.

alexcrichton: It seems that the purpose of this patch is to reduce syscalls setting mode bits, but it only supports Windows, so it seems like a good place to me to add this for Unix? I don't really know how this would interact in a bad way with umask that isn't already exposed, and it seems like it would be simpler to basically remove the `set_perms` call below entirely.

rbtcollins: I'm reasonably sure that chmod(2) which is being used today ignores umask, because its job is setting an explicit mode. But putting that aside.

https://linux.die.net/man/2/chmod suggests to me that we'll get bug reports doing this in creat on Unix (and may even need to tweak this further from doing it at createfile on windows to doing it as a windows specific call on the file handle (current setting readonly requires a stat-set cycle due to the stdlib abstractions).

The specific issue I see here is "On NFS file systems, restricting the permissions will immediately influence already open files, because the access control is done on the server, but open files are maintained by the client. Widening the permissions may be delayed for other clients if attribute caching is enabled on them. " - so untarring a file on NFS will result in the restricted mode being applied on each write and thus erroring.

I'm happy to rework this patch to use winapi to do the readonly setting on the open file without the stat syscalls and avoid this now-identified issue - I don't believe it will affect SMB shares, as SMB's server semantics are much more session oriented, but Windows does have NFS clients.

rbtcollins: Oh, its also worth noting that linux's (and unix in general) decoupled model where IO is permitted to happen long after the client requesting it has gone admits much faster syscalls; fchmod() done late to the file as it is today should be supremely fast on linux and I see no need to change it. I can strace a rustup on linux and see if the std library is causing a lot of wasted syscalls there, and if so tune it as well - but because we don't have to do a metadata call, I don't think that that will be the case.

tl;dr: Unix *already* has less syscalls, and suffers an order of magnitude less from having them in the first place.

alexcrichton: Hm ok, I'm not too particularly intersted in trying to dig into all the dark corners of Unix to see what's going on here. I would personally just like a consistent code path amongst all platforms, and both Unix/Windows support setting file modes on creation so it seems like we should use that. If someone runs into weird issues with weird filesystems later it can be handled, but before that I think internal consistency to ensure that this already very complicated wad of code is as simple as possible is better.

rbtcollins: Ok, so *this* patch isn't going to meet your constraints; lets design a new one. The current code path is
create
write
set mtime on file handle
set mode on file handle
set xattr on file handle

setting the mode on the file handle is fine on unix - its a single syscall per file, and I'm pretty sure that as long as you have decent amounts of RAM available it won't block in any significant fashion.

On windows however, setting the mode looks like this:
```
                    let mut perm = f.metadata()?.permissions();
                    perm.set_readonly(mode & 0o200 != 0o200);
                    f.set_permissions(perm)
```

There are two performance considerations here.
One is that the set_permission syscall is actually the same one that setting the mtime uses: if we abstracted things a little differently we could make a single syscall to set the time and the readonly attribute. A small os-abstraction layer in tar-rs would let us present that cleanly I think; possibly filetime could grow the right interface, or we'd only use filetime for non-windows-family-os's.

Two is that f.metadata()? does two syscalls (I know its written as just GetFileInformationByHandle, but this is the strace: 

<details><summary>strace</summary>
<p>

```

11:33:59.0244734 AM	0.0001502	rustup.exe	25684	CreateFile	C:\Users\robertc\.rustup\tmp\d__ktssiid9tznn8_dir\rust-docs\share\doc\rust\html\cargo\book.css	SUCCESS	Desired Access: Generic Write, Read Attributes, Disposition: Create, Options: Synchronous IO Non-Alert, Non-Directory File, Open Reparse Point, Attributes: n/a, ShareMode: Read, Write, Delete, AllocationSize: 0, OpenResult: Created
11:33:59.0246571 AM	0.0000810	rustup.exe	25684	WriteFile	C:\Users\robertc\.rustup\tmp\d__ktssiid9tznn8_dir\rust-docs\share\doc\rust\html\cargo\book.css	SUCCESS	Offset: 0, Length: 8,192, Priority: Normal
11:33:59.0247826 AM	0.0000287	rustup.exe	25684	WriteFile	C:\Users\robertc\.rustup\tmp\d__ktssiid9tznn8_dir\rust-docs\share\doc\rust\html\cargo\book.css	SUCCESS	Offset: 8,192, Length: 8,192, Priority: Normal
11:33:59.0248378 AM	0.0000088	rustup.exe	25684	WriteFile	C:\Users\robertc\.rustup\tmp\d__ktssiid9tznn8_dir\rust-docs\share\doc\rust\html\cargo\book.css	SUCCESS	Offset: 16,384, Length: 8,192
11:33:59.0248782 AM	0.0000229	rustup.exe	25684	WriteFile	C:\Users\robertc\.rustup\tmp\d__ktssiid9tznn8_dir\rust-docs\share\doc\rust\html\cargo\book.css	SUCCESS	Offset: 24,576, Length: 3,885, Priority: Normal
11:33:59.0249274 AM	0.0000036	rustup.exe	25684	QueryInformationVolume	C:\Users\robertc\.rustup\tmp\d__ktssiid9tznn8_dir\rust-docs\share\doc\rust\html\cargo\book.css	SUCCESS	VolumeCreationTime: 4/01/2019 2:14:43 PM, VolumeSerialNumber: DE41-BA52, SupportsObjects: True, VolumeLabel: 
11:33:59.0249425 AM	0.0000440	rustup.exe	25684	QueryAllInformationFile	C:\Users\robertc\.rustup\tmp\d__ktssiid9tznn8_dir\rust-docs\share\doc\rust\html\cargo\book.css	BUFFER OVERFLOW	CreationTime: 12/05/2019 11:33:59 AM, LastAccessTime: 12/05/2019 11:33:59 AM, LastWriteTime: 12/05/2019 11:33:59 AM, ChangeTime: 12/05/2019 11:33:59 AM, FileAttributes: A, AllocationSize: 49,152, EndOfFile: 28,461, NumberOfLinks: 1, DeletePending: False, Directory: False, IndexNumber: 0x2f00000007eeff, EaSize: 0, Access: Generic Write, Read Attributes, Position: 28,461, Mode: Synchronous IO Non-Alert, AlignmentRequirement: Long
11:33:59.0250234 AM	0.0000123	rustup.exe	25684	SetBasicInformationFile	C:\Users\robertc\.rustup\tmp\d__ktssiid9tznn8_dir\rust-docs\share\doc\rust\html\cargo\book.css	SUCCESS	CreationTime: 1/01/1601 12:00:00 PM, LastAccessTime: 1/01/1601 12:00:00 PM, LastWriteTime: 1/01/1601 12:00:00 PM, ChangeTime: 1/01/1601 12:00:00 PM, FileAttributes: A
11:33:59.0250541 AM	0.0006331	rustup.exe	25684	CloseFile	C:\Users\robertc\.rustup\tmp\d__ktssiid9tznn8_dir\rust-docs\share\doc\rust\html\cargo\book.css	SUCCESS	
```
</p>
</details>

or about 5 hundred-thousands of a second; so for 20k files, thats about 1 second of total runtime, assuming no marshalling overhead etc. As with Unix, we just created the file, we know the attributes we want, so in principle we can set them directly without the intervening file_attr() call; again this could be hidden from the entry() flow to keep that clean and understandable, via a small os module.

But pragmatically; we don't need all of this for rustup. What we need for rustup is a lot simpler, and thats just to stop making the syscalls; as we don't have readonly files there, I'll put forward a SUPER minimal patch and see what you think :).

alexcrichton: Mk so now that we've got the minimal for rustup in I'm actually wondering if it's perhaps best to leave as-is for now until it's an issue? I suspect that large tarballs of readonly files is probably relatively rare. 

In your strace though, do you know what's up with the exponentially growing write sizes? Shouldn't rustup write everything in one go?

rbtcollins: The write lengths are 8K each time, the offsets are 0, 8K, 16K, 24K.  The durations are all over the place - 8,3,1,2 (*10^-6 each).

To write everything in one go, rustup needs to hook into the file output being done by tar-rs - currently thats not possible (at least as far as I understand rust), as tar-rs does the open directly, not through any trait object; and, there is no callback to permit rustup to add an additional BufWriter or similar object around the the File.

But I wouldn't rush to add that just yet; I've done some basic experiments, and at least for rustup, so many of the files are very small - its 20k files, 200M of content, average of 10k per file - that they are already a single write; the small number of quite large files like the js index help drive the common case below 8k. Adding a BufWriter actually slowed things down due to heap allocations. 
I have experimented with patching the stdlib to permit using a larger io::copy buf size but I'm not happy with the outcomes there yet.

I'm happy closing this ticket for now. It's certainly not an itch I have any personal stake in scratching. I agree that very large tars of readonly files in performance-sensitive scenarios is probably quite the edge case.

rbtcollins:
Closing this per discussion - the next tweak if large tars with many  readonly files needs optimising on windows would be to look at a strace of the tar in question and go from there.

PR #203, opened 2019-05-12
Return the File instance from Entry::unpack

rbtcollins:
This permits optimisation of unpacking on Windows - which is perhaps
overly special cased to include in tar itself? Doing this for rustup
which already doesn't use the Archive wrapper :/.

With this patch, I reduced rustup's rust-docs install from 21/22s to
11s on my bench machine - details at
rust-lang/rustup.rs#1850

I haven't filled in all the possibilities for the Unpacked enum, but
happy to do so if you'd like that done - it seemed like it wouldn't
be useful at this stage as there are no live objects to return.

Follow on patch from https://github.com/alexcrichton/tar-rs/pull/202 - again, text conflicts or I would have made them standalone.

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -47,6 +47,18 @@ pub enum EntryIo<'a> {
     Data(io::Take<&'a ArchiveInner<Read + 'a>>),
 }
 
+/// When unpacking items the unpacked thing is returned to allow custom
+/// additional handling by users. Today the File is returned, in future
+/// the enum may be extended with kinds for links, directories etc.
+#[derive(Debug)]
+pub enum Unpacked {
+    /// A file was unpacked.
+    File(std::fs::File),
+    /// A directory, hardlink, symlink, or other node was unpacked.
+    #[doc(hidden)]
+    __Nonexhaustive,
+}
+
 impl<'a, R: Read> Entry<'a, R> {
     /// Returns the path name for this entry.
     ///
@@ -178,7 +190,7 @@ impl<'a, R: Read> Entry<'a, R> {
     ///     file.unpack(format!("file-{}", i)).unwrap();
     /// }
     /// ```
-    pub fn unpack<P: AsRef<Path>>(&mut self, dst: P) -> io::Result<()> {
+    pub fn unpack<P: AsRef<Path>>(&mut self, dst: P) -> io::Result<Unpacked> {
         self.fields.unpack(None, dst.as_ref())
     }
 
@@ -413,14 +425,15 @@ impl<'a> EntryFields<'a> {
     }
 
     /// Returns access to the header of this entry in the archive.
-    fn unpack(&mut self, target_base: Option<&Path>, dst: &Path) -> io::Result<()> {
+    fn unpack(&mut self, target_base: Option<&Path>, dst: &Path) -> io::Result<Unpacked> {
         let kind = self.header.entry_type();
 
         if kind.is_dir() {
-            return self
-                .unpack_dir(dst)
-                .and_then(|_| self.header.mode())
-                .and_then(|mode| set_perms(dst, None, mode, self.preserve_permissions));
+            self.unpack_dir(dst)?;
+            if let Ok(mode) = self.header.mode() {
+                set_perms(dst, None, mode, self.preserve_permissions)?;
+            }
+            return Ok(Unpacked::__Nonexhaustive);
         } else if kind.is_hard_link() || kind.is_symlink() {
             let src = match self.link_name()? {
                 Some(name) => name,
@@ -439,7 +452,7 @@ impl<'a> EntryFields<'a> {
                 )));
             }
 
-            return if kind.is_hard_link() {
+            if kind.is_hard_link() {
                 let link_src = match target_base {
                     // If we're unpacking within a directory then ensure that
                     // the destination of this hard link is both present and
@@ -469,7 +482,7 @@ impl<'a> EntryFields<'a> {
                             dst.display()
                         ),
                     )
-                })
+                })?;
             } else {
                 symlink(&src, dst).map_err(|err| {
                     Error::new(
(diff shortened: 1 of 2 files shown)


src/entry.rs
```diff
@@ -49,6 +49,16 @@ pub enum EntryIo<'a> {
...
 }
 
+/// When unpacking items the unpacked thing is returned to allow custom
+///  additional handling by users.
+#[derive(Debug)]
+pub enum Unpacked {
+    /// A file was unpacked.
+    File(std::fs::File),
+    /// A directory, hardlink, symlink, or other node was unpacked.
+    Other,
```
alexcrichton: Could this perhaps be a nonexhaustive enum? Basically rename this to `__Nonexhaustive` and put `#[doc(hidden)]` on it with other docs indicating that it may be extended in the future. It seems like we'd probably want to extend this in the future to indicate directories that were opened, hard links created, etc.

src/entry.rs
```diff
@@ -594,27 +633,59 @@ impl<'a> EntryFields<'a> {
...
+            mode: u32,
+            _preserve: bool,
+        ) -> io::Result<()> {
+            match f {
+                Some(_f) => {
+                    // Set at file creation
+                    Ok(())
+                }
+                None => {
+                    let mut perm = r#try!(fs::metadata(dst)).permissions();
```
lzutao: ?

src/entry.rs
```diff
@@ -413,14 +425,25 @@ impl<'a> EntryFields<'a> {
...
-            return self
-                .unpack_dir(dst)
-                .and_then(|_| self.header.mode())
-                .and_then(|mode| set_perms(dst, mode, self.preserve_permissions));
+            self.unpack_dir(dst)?;
+            if let Ok(mode) = self.header.mode() {
+                set_perms(dst, None, mode, self.preserve_permissions).map_err(|e| {
+                    TarError::new(
+                        &format!(
+                            "failed to set permissions to {:o} \
```
lzutao: I don't think we should split the line here.

src/entry.rs
```diff
@@ -525,17 +551,29 @@ impl<'a> EntryFields<'a> {
...
+        let mode = self.header.mode();
+        fn open(dst: &Path, _mode: &io::Result<u32>) -> io::Result<std::fs::File> {
+            let mut options = OpenOptions::new();
+            #[cfg(windows)]
+            {
+                // On windows create the file with the desired attributes rather
+                // than setting them in a separate syscall.
+                if readonly(_mode) {
+                    // File is requested to be readonly
+                    // 0x1 is FILE_ATTRIBUTE_READONLY in the Win32 API
```
lzutao: Why not make it a const instead of a comment here?

rbtcollins:
> 
> 
> Seems reasonable to me and looks like some good wins!

Cool. I have some further things I'm poking at - such as being able to set the content indexing disabled flag - this would need to let folk hook into the open() call more deeply, or would need a more complex way of telling unpack what to do; if you have API suggestions around this that would be great.

> This is technically a breaking change but that's ok. I'm probably gonna try to release this as a patch version and pray it doesn't break anything and if that doesn't work out a new major version could be published.

FWIW rustup didn't fail to build with this change added to it - the existing signature io::Result<()> is already must-consume, with the value ignored; as you say "technically breaking" - I too hope no fallout happens.
 
> I expected some further speedups on behalf of rustup from removing extra syscalls on the extracted files (using the returned handle instead of reopening), but I was surprised to see that the main purpose was actually just dropping files! Isn't this a case where if extraction happened in parallel on many threads it'd also solve the issue?

Yes, if extraction was happening in parallel on many threads, CloseHandle would be per-thread as a consequence. However handling of dependencies in tar files becomes more complex with parallel extraction (e.g. extract foo/ then foo/bar ). This by comparison is very simple to reason about (see the patch in rustup).

Another thing we can parallelise for more speed is the decompression of the tar content; and then also another person is interested in extracting while downloading from the net for even better wall clock times.

src/entry.rs
rbtcollins: This is code from #201

src/entry.rs
rbtcollins: This is identical to the hunk in line 631; I'd like to leave them identical and do a follow up patch to make error formatting / capturing more consistent - I think its had variation creep in over time, and we can push some of the error formatting into helpers to make it more consistent.

src/entry.rs
rbtcollins: Thats from #202

src/entry.rs
rbtcollins: I think I've done this right.

src/entry.rs
alexcrichton: Can the deduplication be removed?

rbtcollins: sure; was offering to keep it as a separate patchset, will push up into this branch shortly

rbtcollins: I've factored that into an intermediate function; let me know if you want it done differently etc. I've also removed this from layering on #202 as that is looking more complex - and they no longer conflict - so this should be mergable now (if you're happy with the code).

Without that I only get down to 12s unpacks, not 9/10s but still, its a significant win.

src/entry.rs
```diff
@@ -504,14 +518,16 @@ impl<'a> EntryFields<'a> {
...
+            return Ok(Unpacked::__Nonexhaustive);
         };
 
         // Old BSD-tar compatibility.
         // Names that have a trailing slash should be treated as a directory.
         // Only applies to old headers.
         if self.header.as_ustar().is_none() && self.path_bytes().ends_with(b"/") {
-            return self.unpack_dir(dst);
+            self.unpack_dir(dst)?;
+            // Why no permission setting?
```
alexcrichton: This is probably just a bug and you can feel free to update this

rbtcollins: done

PR #204, opened 2019-05-16
Bug #198: Set mtime and atime on file handles

rbtcollins:
For unix systems this should be marginally more efficient; for
Windows systems a whole open/close pair is avoided, which can make
significant savings when dealing with many files.


rbtcollins:
This will fix the test failure in #202

alexcrichton:
👍

PR #205, opened 2019-05-17
Avoid unneeded readonly syscalls on Windows

rbtcollins:
This meets the rustup use case (less syscalls) but not the general use case (unpacking a huge tar with many readonly files will be slower than a huge tar with many writable files)

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -620,15 +620,18 @@ impl<'a> EntryFields<'a> {
             mode: u32,
             _preserve: bool,
         ) -> io::Result<()> {
+            if mode & 0o200 == 0o200 {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Ah ok this definitely makes sense to me, thanks for this! We can continue to work on the other patch if you'd like as well for handling tarballs with lots of readonly files!

rbtcollins:
The other patch is a lower pri for me but I'm happy to put some effort into it out of general principle that things should be nice ;). Could we cut a release of tar-rs at this point though, to get the improvements thus far into rustup ?

alexcrichton:
Sure thing, done!

PR #206, opened 2019-05-17
Release 0.5

rbtcollins:
Technically a breaking change due to Entry::unpack changing signature
but not expected to break anyone.


rbtcollins:
ha, raced with you.

PR #215, opened 2019-11-08
Migrate tempdir dependency to tempfile

silwol:
The tempdir crate is deprecated and tempfile is recommended as a
successor by its developers.


alexcrichton:
Thanks!

PR #216, opened 2019-11-10
Fix Windows ignoring preserve_permissions

Jake-Shadle:
If you use `set_preserve_permissions` and set to `false`, Windows would completely ignore this and continuing adding the read-only flag.

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -645,9 +645,9 @@ impl<'a> EntryFields<'a> {
             dst: &Path,
             f: Option<&mut std::fs::File>,
             mode: u32,
-            _preserve: bool,
+            preserve: bool,
         ) -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! This appears, however, to not be what the `preserve_permissions` flag is used for. That's just for applying `0o777` as a mask to the listed mode on Unix, but even on Unix it still actually executes `chmod` to change the permissions. To avoid the syscall altogether, perhaps a new flag is necessary?

Jake-Shadle:
Hmm, I guess I misinterpreted what it was for, and I realized fixing a bug that it was the root cause of the issue I was seeing, not tar setting the read-only flag on directories, so I'll just close this for now.

rbtcollins:
See also https://github.com/alexcrichton/tar-rs/pull/205

PR #220, opened 2019-12-11
Allow current directories in header paths

### LINKED ISSUE #219: Don't remove CurrentDir on Copy Path
The current logic for copy `copy_path_into` will remove any current directories, aka `.`, from the path. For example, if you pass in the path `./control` the header will have the path `control`.

This logic can be found here:

https://github.com/alexcrichton/tar-rs/blob/8b141d0d7119a412373a558d0b884c519b6d44a8/src/header.rs#L1497-L1500

I'm trying to use this repository to build Debian packages. Those packages contain a control.tar file which contain a file named `./control`. With the current implementation of tar-rs, its impossible to mimic this functionality.

My suggestion would be to not remove any current directories in `copy_path_into`.

schultetwin1:
The current implementation removes any "current directories" from the
header path for a file. This change stops the removal to leave the path
as is.

This change adds a test to check for regression

Fixes #219

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -362,7 +362,8 @@ impl Header {
     ///
     /// This function will set the pathname listed in this header, encoding it
     /// in the appropriate format. May fail if the path is too long or if the
-    /// path specified is not Unicode and this is a Windows platform.
+    /// path specified is not Unicode and this is a Windows platform. Will
+    /// strip out any "." path component, which signifies the current directory.
     pub fn set_path<P: AsRef<Path>>(&mut self, p: P) -> io::Result<()> {
         self._set_path(p.as_ref())
     }
@@ -414,7 +415,8 @@ impl Header {
     ///
     /// This function will set the linkname listed in this header, encoding it
     /// in the appropriate format. May fail if the link name is too long or if
-    /// the path specified is not Unicode and this is a Windows platform.
+    /// the path specified is not Unicode and this is a Windows platform. Will
+    /// strip out any "." path component, which signifies the current directory.
     pub fn set_link_name<P: AsRef<Path>>(&mut self, p: P) -> io::Result<()> {
         self._set_link_name(p.as_ref())
     }
--- a/tests/header/mod.rs
+++ b/tests/header/mod.rs
@@ -139,6 +139,12 @@ fn set_path() {
         assert_eq!(t!(h.path()).to_str(), Some("foo\\bar"));
     }
 
+    // set_path documentation explictly states it removes any ".", signfying the
+    // current directory, from the path. This test ensures that documented
+    // beavhior occurs
+    t!(h.set_path("./control"));
+    assert_eq!(t!(h.path()).to_str(), Some("control"));
+
     let long_name = iter::repeat("foo").take(100).collect::<String>();
     let medium1 = iter::repeat("foo").take(52).collect::<String>();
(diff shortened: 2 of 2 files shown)


alexcrichton:
Thanks for the report and PR! Could you elaborate though to the motivation for this? I'm not sure why it would be necessary to keep the `.` directives in there

schultetwin1:
For sure! The issue is paths passed into `tar::Builder::append_data` (or some of the other `append` functions) are not comparable to what is returned from `tar::Header::path`. The following code ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b06f685a819dbd635787541fe99895f0)) is a basic example of this:

```rust
let mut header = Header::new_gnu();
let path = std::path::Path::new("./control");
header.set_path(&path).unwrap();
    
println!("Path: {}", path.display());
println!("Path in Header: {}", header.path().unwrap().display());
assert!(path == header.path().unwrap());
```

This will print

> Path: ./control
> Path in Header: control

and then panic on the assertion. This makes it hard to iterate through a tar to find a file you put in it.

`./control` and `control` are technically the same path but I don't know how to detect that using `std::path::Path`. It looks like there is some work for support to normalize paths but it does not exists yet. https://github.com/rust-lang/rfcs/issues/2208

Maybe another option would be to provide a function which normalizes paths for the user so they can see how the path they passed into the crate might have changed?

alexcrichton:
Ah ok, I see!

Would counting files work alright for your use case? Indexing files by their position in the archive should be consisten across implementations

schultetwin1:
Unfortunately, counting files will not work for me. There is no defined order for the files in the tar in my use case (more info below if you are curious).

I have a simple work around to get past this issue, I was just filing this issue and this fix so that others do not run into this same issue.  It took me a little while to figure out that tar-rs was silently swallowing the `./` in my path.

If you would prefer to not make this change, I'd like to switch this PR to document the behavior of the `append` functions modifying the `path` parameter the user passed in to not including the current directory `.`.

For more background:

I'm trying to create and parse [binary debian packages](https://www.debian.org/doc/debian-policy/ch-binary.html). The metadata for a binary debian package is stored in a file named "control"  which in turn is inside a archive (control.tar).

In most debian packages I have found in ubuntu's package repository the "control" file has a path of `./control` in their tar file. So I was trying to mimic that behavior but it's impossible to with this library.

alexcrichton:
Thanks for the info! And yeah if you're ok documenting this for now, that seems good to me to add!

schultetwin1:
Sounds good!

For my own learning, why do you want to keep in this path normalization logic which strips paths of the current directory path? Is there some edge case related to the tar file format that we can avoid by doing this?

alexcrichton:
Oh I'm not really particularly wed to the logic one way or another, I was mostly curious as to the motivations for this. Some degree of path normalization is done to avoid dealing with `..` in names or giving errors when it goes outside something. Other path normalization is required to translate `\` to `/` on Windows. Ignoring `.` for the current directory just sort of fell out of all of this because it was already happening and was easy enough to handle.

I don't think that anything would break with this but I'm not 100% confident myself, hence the question for motivation to see how deeply I needed to look!

schultetwin1:
My apologies for the delay. Updated this change to be documentation only.

alexcrichton:
👍

PR #224, opened 2020-05-19
Add case for ustar header with gnu longname

### LINKED ISSUE #223: Fails to unpack long paths
On Windows `tar-rs` fails to unpack an archive when the archive contains a folder containing a file where the name of the folder is long enough to require a `@LongLink`. See here for a [repo ](https://github.com/davidMcneil/tar-long-path-repro) replicating the issue.

Essentially, unpacking an archive with a directory structure as shown below fails.
* `this_is_a_file_with_a_name_that_just_keeps_going_on_and_on_and_on_it_is_a_very_very_long_file_name`
    * `test.txt`

Code used to unpack:

```rust
use tar::Archive;
use std::fs::File;

fn main() {
    let file = File::open("test/test.tar").expect("failed to open tar file");
    let mut tar = Archive::new(file);
    tar.unpack("test/expanded/").expect("failed to unpack tar file");
}
```

The underlying error is "Access is denied." By adding logging to tar you can see the order of entries processed is as follows:

    "test/expanded/this_is_a_file_with_a_name_that_just_keeps_going_on_and_on_and_on_it_is_a_very_very_long_file_name" Directory
    "test/expanded/@LongLink" GNULongName
    "test/expanded/this_is_a_file_with_a_name_that_just_keeps_going_on_and_on_and_on_it_is_a_very_very_long_file_name" Regular

The third entry is pointing to the directory when it should be pointing to the file. The "Access is denied." error is the result of trying to open a directory as a file.

davidMcneil:
Resolves #223 

When using `7zip` to make an archive on Windows it creates an archive with a `ustar` header with entries of type `GnuLongname`. This PR adds support for this case.

I know very little about the `tar` spec. It would be great if someone could confirm that this is the correct fix. Should support also be added for `ustar` with `GnuLonglink`?

Signed-off-by: David McNeil <mcneil.david2@gmail.com>


alexcrichton:
Thanks for the PR! Is this perhaps a new feature in 7zip? Would it be possible to file an issue with them indicating that it may be buggy?

alexcrichton:
Or alternatively, if there's precedent of doing this in other libraries, then we could do that.

davidMcneil:
> Is this perhaps a new feature in 7zip?

I do not believe so. I did some more investigation. This bug also exists for archives created on linux with `7zip` version 16.02 from 2016-05-21. Here are the steps to create a failing archive:

```
mkdir this_is_a_folder_with_a_name_that_just_keeps_going_on_and_on_and_on_it_is_a_very_very_long_folder_name
touch this_is_a_folder_with_a_name_that_just_keeps_going_on_and_on_and_on_it_is_a_very_very_long_folder_name/test.txt
7z a -ttar test.tar this_is_a_folder_with_a_name_that_just_keeps_going_on_and_on_and_on_it_is_a_very_very_long_folder_name/
tar x -f ../test.tar

```

`tar x -f 7z_long_path.tar` works fine so I believe this is a valid archive.

> Or alternatively, if there's precedent of doing this in other libraries, then we could do that.

I can do some investigation. What projects would you recommend looking at? `libarchive`?

davidMcneil:
Is it possible the checks on the header type ([1](https://github.com/alexcrichton/tar-rs/blob/8e3a9de0287de7795684c8f22eaaf1f7e2c62d47/src/archive.rs#L290), [2](https://github.com/alexcrichton/tar-rs/blob/8e3a9de0287de7795684c8f22eaaf1f7e2c62d47/src/archive.rs#L301), [3](https://github.com/alexcrichton/tar-rs/blob/8e3a9de0287de7795684c8f22eaaf1f7e2c62d47/src/archive.rs#L312)) should be removed (ie do not check for gnu vs ustar) and only use the `entry_type` in the if statements.

davidMcneil:
It looks like `libsodium` treats the [two formats identically](https://github.com/libarchive/libarchive/blob/df2cd8c4687d8384ad6ccc16ad5fe54428abdef2/libarchive/archive_read_support_format_tar.c#L390) identically which maybe supports the removal of the `gnu` vs `ustar` header checks as commented above.

alexcrichton:
Ah ok that's good enough for me! In that case want to remove the check for `is_gnu` above and that should be good to go?

alexcrichton:
(or actually, update it to `is_gnu || is_ustar`

davidMcneil:
I updated the PR to check for both header formats.

alexcrichton:
Thanks!

PR #225, opened 2020-05-21
Remove special casing for Redox

jackpot51:
Redox is part of the unix target family now and supports all the libc functions used in tar-rs


jackpot51:
Thanks @alexcrichton, could you make a new release?

PR #228, opened 2020-06-01
Add case for entry size in the pax extension

irisliu88:
This commit adds support for when the size of the entry does not reside
in the block header and instead with the associated local pax extension.
We check if there is a 'size' keyword present in the extension and use its
value as the entry size if the existing header size is zero.

Signed-off-by: Iris Liu <iris@onechronos.com>


src/archive.rs
```diff
@@ -332,6 +338,25 @@ impl<'a> EntriesFields<'a> {
         }
     }
 
+    fn get_pax_extension_size(&mut self, pax: &[u8]) -> Option<u64> {
+        for extension in pax_extensions(pax) {
+            let current_extension = extension.unwrap();
```
alexcrichton, line 343: The `.unwrap()` here should be avoided, this should likely just break out of the loop

src/archive.rs
```diff
@@ -332,6 +338,25 @@ impl<'a> EntriesFields<'a> {
...
+            let value = match current_extension.value() {
+                Ok(value) => value,
+                Err(_) => return None,
+            };
+            let key = match current_extension.key() {
+                Ok(key) => key,
+                Err(_) => return None,
+            };
+            if key == "size" {
```
alexcrichton, line 352: This might be a bit simpler as:

```rust
if current_extension.key() != Some("size") {
    continue;
}
// ...
```

src/archive.rs
```diff
@@ -332,6 +338,25 @@ impl<'a> EntriesFields<'a> {
...
+                let size = value.parse::<u64>().unwrap();
```
alexcrichton, line 353: This `.unwrap()` should be handled by skipping the entry or returning `None`

irisliu88:
> Thanks! In terms of organization could this avoid storing a `size` field in the `EntryFields`? Since that's not as simple as it used to be it might be best to have the calculation of pax/etc done in `fn size(&self)`.

irisliu88:
> Thanks! In terms of organization could this avoid storing a `size` field in the `EntryFields`? Since that's not as simple as it used to be it might be best to have the calculation of pax/etc done in `fn size(&self)`.

I see, so do you mean to keep the size field zero in this case still?

alexcrichton:
In the header since it's reading raw values that's what it would return, yeah, but `Entry::size` would still be added as a method and the calculation would be done there to read the pax extensions.

src/archive.rs
```diff
@@ -241,10 +243,12 @@ impl<'a> EntriesFields<'a> {
...
         let file_pos = self.next;
-        let size = header.entry_size()?;
-
+        let mut size = header.entry_size()?;
+        if (size == 0) && (pax_size.is_some()) {
+            size = pax_size.unwrap();
+        }
         let ret = EntryFields {
-            size: size,
+            size: header.entry_size()?,
```
alexcrichton, line 251: Could this field be removed entirely? That way there's no need to keep it in sync, and users can use the `size()` method

src/archive.rs
```diff
@@ -241,10 +243,11 @@ impl<'a> EntriesFields<'a> {
         }
 
         let file_pos = self.next;
-        let size = header.entry_size()?;
-
+        let mut size = header.entry_size()?;
+        if (size == 0) && (pax_size.is_some()) {
+            size = pax_size.unwrap();
+        }
```
alexcrichton, line 249: I think this is dead code now?

src/entry.rs
```diff
@@ -271,7 +283,7 @@ impl<'a> EntryFields<'a> {
 
     pub fn read_all(&mut self) -> io::Result<Vec<u8>> {
         // Preallocate some data but don't let ourselves get too crazy now.
-        let cap = cmp::min(self.size, 128 * 1024);
+        let cap = cmp::min(self.header.entry_size()?, 128 * 1024);
```
alexcrichton, line 286: Presumably this should use `self.size()?`, right?

alexcrichton:
Er hm actually I'm getting very mixed up as to what's going on... There's lots of places that are checking this and sometimes seem to consult the pax header vs the real size header. Are you sure that this is correctly handled for previous archives as well as archives with this pax header?

src/archive.rs
irisliu88: `size` is needed to advance the `self.next` pointer

irisliu88: Also this governs the `data` size in `EntryFields`

alexcrichton:
Ok sorry for taking a moment to get back to this. My worry is still that the pax size is only consulted in some locations and not others. I think, for example `read_all` only consults the header size (zero?) instead of the pax size. Additionally in other locations like `parse_sparse_header` it's using `entry_size()` which isn't taking into account the pax size.

Is there a reason that the pax size is only considered in a few locations and not others?

irisliu88:
> Ok sorry for taking a moment to get back to this. My worry is still that the pax size is only consulted in some locations and not others. I think, for example `read_all` only consults the header size (zero?) instead of the pax size. Additionally in other locations like `parse_sparse_header` it's using `entry_size()` which isn't taking into account the pax size.
> 
> Is there a reason that the pax size is only considered in a few locations and not others?

In order for all internal `header.entry_size()` to account for the pax size, I've moved the function to reside within `EntryFields`. Let me know if I've missed any other areas. I've also renamed the original `Entry` `size()` to be `entry_size()` as there is a difference for sparse headers.

src/archive.rs
```diff
@@ -425,7 +430,7 @@ impl<'a> EntriesFields<'a> {
                  size in header",
             ));
         }
-        entry.size = cur;
+        entry.header.set_size(cur);
```
alexcrichton, line 433: I sort of forget where this is all called from, but this `set_size` feels not quite right here because it's rewriting the header that users may be inspecting. I think it would be best if this `cur` is still saved off in a `tar`-crate-specific location.

src/archive.rs
```diff
@@ -320,6 +324,7 @@ impl<'a> EntriesFields<'a> {
                     ));
                 }
                 pax_extensions = Some(EntryFields::from(entry).read_all()?);
+                pax_size = pax_extensions_size(&pax_extensions.as_ref().unwrap());
```
alexcrichton, line 327: Could the `.unwrap()` be avoided here by parsing before storing the raw data in the `pax_extensions` local?

src/archive.rs
```diff
@@ -241,10 +243,11 @@ impl<'a> EntriesFields<'a> {
         }
 
         let file_pos = self.next;
-        let size = header.entry_size()?;
-
+        let mut size = header.entry_size()?;
+        if (size == 0) && (pax_size.is_some()) {
+            size = pax_size.unwrap();
```
alexcrichton, line 248: Can this `.unwrap()` be avoided with `if let`?

src/archive.rs
irisliu88: This is called at the end of reading `next_entry()`. It was was originally stored at the `EntryFields` `size` field, but since that is no longer present, where is the best location given we do not want to create any additional fields?

alexcrichton: It may be the case that my initial inclination to not duplicate the information in a few places may not have been accurate. It may be that we need to bring back a `size` field in the struct

irisliu88: Hmm, yeah I think so. I'll make the changes this week.

irisliu88: I've reverted the changes back to using the same `size` field. I do think it is cleanest currently while the sparse files need to modify the `size`.

alexcrichton:
Ok thanks again for being patient with this!

PR #230, opened 2020-06-11
cannonicalize unpack destination to safely handle long paths on windows

mwrock:
This addresses a scenario on windows where files may be unpacked with a path over 260 (`MAX_PATH`) charachters. Such a case results in a `NotFound` error when creating the file. This cannonicalizes the unpack `dst` which will prepend `\\?\` to the path allowing Windows APIs to treat the path as an extended-length path supporting 32,767 characters.

I can understand if the author of this crate would prefer callers to cannonicalize the path themselves. However in the process of migrating from libarchive to this crate, this behavior was surprising since libarchive [performs the prepending](https://github.com/libarchive/libarchive/blob/3649ed23c6b4392d692580c03b10a611e3eaaa32/tar/bsdtar_windows.c#L61) of `\\?\` for the caller.

Signed-off-by: mwrock <matt@mattwrock.com>

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -1,5 +1,6 @@
 use std::cell::{Cell, RefCell};
 use std::cmp;
+use std::fs;
 use std::io;
 use std::io::prelude::*;
 use std::marker;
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for this!

I think it's reasonable to add this behavior, especially if it's already done by other implementations. That being said though I think I'd prefer to have the behavior tweaked here a bit with a few points:

* If `canonicalize` fails, can the previous input be used instead? That way even if it fails we'll still try and see if there's a different failure later.
* I'm a bit worried about the cost of this about how it's canonicalizes once-per-entry. Would it be possible to instead canonicalize once at the request for extraction of an entire tarball? Otherwise this could also have a check to see if, on Windows, it's already in a "canonical form" that subverts the path length limits and skip the canonicalize in that case.

mwrock:
Sorry I missed how `unpack_in` was being called per entry. That would indeed have been costly. I just pushed up code to address both your concerns @alexcrichton . Thanks!

alexcrichton:
Thanks! Mind also adding a test for this for CI?

Also, is the creation of the directory related to fixing this for Windows? Or is it trying to keep behavior in-line with other tools?

mwrock:
Will add a test. Canonicalize only works for an existing path, otherwise it throws an error.  This path would be [created down the road](https://github.com/alexcrichton/tar-rs/blob/87bc5169c63acccb3650e5fb5dbeb8039e529686/src/entry.rs#L396) but creating the dst base here makes the canonicalize  possible.

mwrock:
test added @alexcrichton

alexcrichton:
👍

PR #233, opened 2020-06-15
Add documentation on path being directory

### LINKED ISSUE #231: Clear documentation of append_dir() / append_path() about recursiveness
The functions `append_dir` and `append_path`, add `path` to the archive if it's a `file`, otherwise, these functions add an entry if it's a `dir`. I think it should be mentioned in the documentation. 

Only `append_dir_all` adds contents recursively. It is mentioned.

Arkanayan:
Document behaviour in case the input path is a directory.
resolves #231


alexcrichton:
Thanks!

PR #234, opened 2020-06-17
Add contiguous mode to the Builder

### LINKED ISSUE #227: Add a way to use Builder in append mode, that would not write empty blocks on the end of the archive
I've very recently used this library to stream tar.gz output using Rocket. While I was able to do this pretty easily, I had to give up using `tar::Builder` only because it forces me to call `finish` in all destructors (e.g.`drop` or `into_inner`).

Imagine, that I have `StreamedBody` struct that implements `Read`. In `read`, it loads some resources (for example from the DB or file system) and archives them using tar. Since it loads only data chunk, it also has to produce a chunk of tar = header + data. There is great `tar::Builder::append_data` method but 

1) `Builder` has to live only inside `Read::read`, otherwise I run into borrowing issues (`Builder` holds a reference to a buffer. In my example, both `Builder` and buffer would be fields of `StreamedBody` which is not easily achievable)
2) because of (1) and because Builder writes empty blocks in `drop`, every time a chunk is archived, the end of archive mark is written.
 
I think we could add `tar::Builder::continuous(&mut self)` that would set a flag and then in `tar::Builder::finish` we could only write empty blocks if continuous == false.

synek317:
Closes #227 

There are many ways to implement this. Here is my proposal, I'm open to any suggestions.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -15,6 +15,7 @@ pub struct Builder<W: Write> {
     mode: HeaderMode,
     follow: bool,
     finished: bool,
+    contiguous: bool,
     obj: Option<W>,
 }
 
@@ -27,6 +28,7 @@ impl<W: Write> Builder<W> {
             mode: HeaderMode::Complete,
             follow: true,
             finished: false,
+            contiguous: false,
             obj: Some(obj),
         }
     }
@@ -42,6 +44,13 @@ impl<W: Write> Builder<W> {
     /// than adding a symlink to the archive. Defaults to true.
     pub fn follow_symlinks(&mut self, follow: bool) {
         self.follow = follow;
+    }
+
+    /// Determines wheter the archive will be automatically finished
+    /// Contiguous Builder will not finish the archive unless `finish`
+    /// is explicitly called. Defaults to false.
+    pub fn contiguous(&mut self, contiguous: bool) {
+        self.contiguous = contiguous;
     }
 
     /// Gets shared reference to the underlying object.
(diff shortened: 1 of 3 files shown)


src/builder.rs
```diff
@@ -356,6 +363,14 @@ impl<W: Write> Builder<W> {
         self.finished = true;
         self.get_mut().write_all(&[0; 1024])
     }
+
+    fn finalize(&mut self) -> io::Result<()> {
+        if !self.contiguous {
```
alexcrichton, line 368: Since `finish()` is a public method, I think that it might be best if this check is folded into there? That can just keep the documentation, though, that it is not required to be called.

synek317: I've introduced `finalize` since I wanted to give the user the possibility to write the 'footer' section, as you can see in the example.

I could move the check to `finish` and then create public method `write_footer_section`. Then this:

    let mut builder = tar::Builder::new(...);
    builder.skip_footer_section(true);
    builder.finish();

would have no effect and would need to be transformed into:

    let mut builder = tar::Builder::new(...);
    builder.skip_footer_section(true);
    builder.write_footer_section();

alexcrichton: I think though that this implementation, as-is, doesn't work. If you call `finish` it ignores the `skip_footer_section` flag introduced here, which I think is a bug. I think it'd be fine to basically say that `finish` is optional if `skip_footer_section` is `true`.

synek317: It does work, please take a look at the example I've added to `examples/` or even tests. But I see your point, I will rename `contiguous` to `skip_footer_section` and move

    self.get_mut().write_all(&[0; 1024]) 

To new public method `write_footer_section`. Would that work in your opinion?

alexcrichton: Yes while it works I think it's too confusing. I think adding a separate method makes sense. That way there's a flag to disable auto-inclusion of the footer and in that scenario you can either create a `tar::Builder` at the end with the flag enabled or you can manually call the footer write.

PR #240, opened 2021-01-25
Remove dead dependency on redox-syscall

benesch:
Usage of the redox-syscall crate was removed in #225, but its entry in
Cargo.toml got left behind.


alexcrichton:
👍

PR #244, opened 2021-02-18
fix: allow unaligned final blocks in sparse archives

### LINKED ISSUE #243: Err on GNU sparse files with content at end
GNU tar `--sparse` produces archives we can't process, we hit this error:
https://github.com/alexcrichton/tar-rs/blob/8d51b538da744eb7a30022b2a3d13b6a49f6f419/src/archive.rs#L410-L413

A sparse file with content at the end, such as:

```
$ for i in $(seq 0 1); do truncate -s ${i}M a.big; echo "${i}MB through" >> a.big; done
$ hexyl a.big
┌────────┬─────────────────────────┬─────────────────────────┬────────┬────────┐
│00000000│ 30 4d 42 20 74 68 72 6f ┊ 75 67 68 0a 00 00 00 00 │0MB thro┊ugh_0000│
│00000010│ 00 00 00 00 00 00 00 00 ┊ 00 00 00 00 00 00 00 00 │00000000┊00000000│
│*       │                         ┊                         │        ┊        │
│00100000│ 31 4d 42 20 74 68 72 6f ┊ 75 67 68 0a             │1MB thro┊ugh_    │
└────────┴─────────────────────────┴─────────────────────────┴────────┴────────┘
```

Packed into a tar with `tar -cf a.tar --sparse a.big` into [a.tar.gz](https://github.com/alexcrichton/tar-rs/files/5970459/a.tar.gz)
 apparently produces this header structure:

```
regular header 0: GnuSparseHeader { offset: 0, length: 4096 }
regular header 1: GnuSparseHeader { offset: 1048576, length: 12 }
regular header 2: GnuSparseHeader { offset: 1048588, length: 0 }
```

...and we hit the error as `12 % 512 != 0`.

I notice `remaining` does hit zero after this subtraction:

https://github.com/alexcrichton/tar-rs/blob/8d51b538da744eb7a30022b2a3d13b6a49f6f419/src/archive.rs#L426-L431

Do we need to break the header parsing if we see a sparse header with a length of zero?
Do we need to skip the 512 byte alignment check for the last chunk, or if we're at the end of the entry?

Tested with GNU tar 1.30 (ubuntu 20.10 amd64). We can parse its `-H pax` output fine.

FauxFaux:
We incorrectly reject archives which do not have 'correct' alignment
on the marker block at the end of the sparse run. Skip this check for
the final (length = 0) block.

No validation is done that this is actually the final block.

Fixes #243.


alexcrichton:
Thanks!

PR #246, opened 2021-04-03
builder: add support for archiving special files on Unix

liushuyu:
Added support for archiving character, block and FIFO files on Unix, this is useful when you want to take a whole system backup or re-archiving an extracted system image.

The file handling in this pull request isn't particularly graceful since I am not very experienced with the codebase, any suggestions are welcome!

---
This change is [Reviewable](https://reviewable.io/reviews/alexcrichton/tar-rs/246)

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -410,6 +410,39 @@ fn append_path_with_name(
             Some(&link_name),
         )
     } else {
+        #[cfg(unix)]
+        {
+            append_special(dst, path, &stat, mode)
+        }
+        #[cfg(not(unix))]
+        {
+            Err(other(&format!("{} has unknown file type", path.display())))
+        }
+    }
+}
+
+#[cfg(unix)]
+fn append_special(
+    dst: &mut dyn Write,
+    path: &Path,
+    stat: &fs::Metadata,
+    mode: HeaderMode,
+) -> io::Result<()> {
+    use ::std::os::unix::fs::FileTypeExt;
+    let file_type = stat.file_type();
+    if file_type.is_socket() {
+        // sockets can't be archived
+        Err(other(&format!(
+            "{}: socket can not be archived",
+            path.display()
+        )))
+    } else if file_type.is_fifo() {
+        append_fs_special(dst, path, &stat, EntryType::Fifo, mode)
+    } else if file_type.is_char_device() {
+        append_fs_special(dst, path, &stat, EntryType::Char, mode)
+    } else if file_type.is_block_device() {
+        append_fs_special(dst, path, &stat, EntryType::Block, mode)
+    } else {
         Err(other(&format!("{} has unknown file type", path.display())))
     }
 }
@@ -518,6 +551,32 @@ fn append_fs(
     append(dst, &header, read)
 }
 
+#[cfg(unix)]
+fn append_fs_special(
+    dst: &mut dyn Write,
+    path: &Path,
+    meta: &fs::Metadata,
+    entry_type: EntryType,
+    mode: HeaderMode,
+) -> io::Result<()> {
+    use ::std::os::unix::fs::MetadataExt;
+    let mut header = Header::new_gnu();
+    header.set_metadata_in_mode(meta, mode);
+    prepare_header_path(dst, &mut header, path)?;
+
+    header.set_entry_type(entry_type);
+    let dev_id = meta.rdev();
+    let dev_major = ((dev_id >> 32) & 0xffff_f000) | ((dev_id >> 8) & 0x0000_0fff);
+    let dev_minor = ((dev_id >> 12) & 0xffff_ff00) | ((dev_id) & 0x0000_00ff);
+    header.set_device_major(dev_major as u32)?;
+    header.set_device_minor(dev_minor as u32)?;
+
+    header.set_cksum();
+    dst.write_all(header.as_bytes())?;
+
+    Ok(())
+}
+
 fn append_dir_all(
     dst: &mut dyn Write,
     path: &Path,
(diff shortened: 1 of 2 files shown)


alexcrichton (commented):
Seems reasonable enough to me, thanks!

Could this perhaps be refactored a bit though to avoid a few extra calls to `fs::metadata` being called? I think this could have an internal append function which takes a `FileType` from libstd and dispatches internally about how to append it? (e.g. more cases would be handled on Unix than on Windows, but regular files would all roughly be handled in the same place)

liushuyu:
> Could this perhaps be refactored a bit though to avoid a few extra calls to `fs::metadata` being called? I think this could have an internal append function which takes a `FileType` from libstd and dispatches internally about how to append it? (e.g. more cases would be handled on Unix than on Windows, but regular files would all roughly be handled in the same place)

Sorry for the late reply. I am a bit confused by this suggestion: I saw the https://github.com/alexcrichton/tar-rs/blob/f07b3c15e398a76a627e77b49952d44c5d5880d8/src/builder.rs#L542 part so I assumed one more `stat()` should be fine here? https://github.com/alexcrichton/tar-rs/pull/246/files#diff-e4f794ca308aa712cbad31feb714df7cada1bbe453a30e232e45891571430263R607

Did you mean to refactor the `append_dir_all` function so that the `stat` could be avoided? Or should I do something about `append_special` function?

alexcrichton:
I mostly just got the feeling reading this that it'd be better to refactor where at some point there's the equivalent of a `match` on the `FileType` (or similar) of what to archive, and right now it seems spread across a few locations with perhaps an extra metadata call here or there.

liushuyu:
I have made an attempt to reduce `stat` calls:

image

src/builder.rs
```diff
@@ -518,22 +551,58 @@ fn append_fs(
...
+        let file_type = metadata.file_type();
+        let is_dir = file_type.is_dir();
+        let is_symlink = file_type.is_symlink();
         // In case of a symlink pointing to a directory, is_dir is false, but src.is_dir() will return true
         if is_dir || (is_symlink && follow && src.is_dir()) {
             for entry in fs::read_dir(&src)? {
                 let entry = entry?;
-                let file_type = entry.file_type()?;
-                stack.push((entry.path(), file_type.is_dir(), file_type.is_symlink()));
+                let metadata = entry.metadata()?;
```
alexcrichton, line 604: This unfortunately has a side effect where on platforms like Unix this ends up doing a syscall whereas previously accessing `file_type` generally doesn't require a syscall on most major platforms. The intention was that this loop should primarily only do `fs::read_dir` as a sycall and otherwise other syscalls can be avoided.

My thinking about refactoring this was largely that this loop has one iteration of "look at the file type and delegate appropriately" but above in `append_path_with_name` it's basically the same loop. With the `append_path_with_name` addition it's getting more complicated so I was wondering if it would be possible to refactor this loop and `append_path_with_name` above to call a common function which is the one location where special files are handled.

src/builder.rs
```diff
@@ -518,22 +551,58 @@ fn append_fs(
     append(dst, &header, read)
 }
 
+#[cfg(unix)]
+fn append_fs_special(
```
alexcrichton, line 555: This is only called by `append_special`, so can it be merged there?

tests/all.rs
```diff
@@ -1149,3 +1149,40 @@ fn append_long_multibyte() {
...
+        }
+    }
+
+    t!(env::set_current_dir(td.path()));
+    let mut ar = Builder::new(Vec::new());
+    // append_path has a different logic for processing files, so we need to test it as well
+    t!(ar.append_path("fifo"));
+    t!(ar.append_dir_all("special", td.path()));
+    // unfortunately, macOS CI can't handle `set_current_dir` correctly
+    if !cfg!(target_os = "linux") {
```
alexcrichton, line 1177: Could the `#[cfg]` on the test be updated to `target_os = "linux"`?

src/builder.rs
```diff
@@ -409,6 +409,39 @@ fn append_path_with_name(
...
+        Err(other(&format!(
+            "{}: socket can not be archived",
+            path.display()
+        )))
+    } else if file_type.is_fifo() {
+        append_fs_special(dst, path, &stat, EntryType::Fifo, mode)
+    } else if file_type.is_char_device() {
+        append_fs_special(dst, path, &stat, EntryType::Char, mode)
+    } else if file_type.is_block_device() {
+        append_fs_special(dst, path, &stat, EntryType::Block, mode)
```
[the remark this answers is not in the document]
liushuyu: Addressed in 7b5fec1f91e961b2fafe6a897b36fd9d06366978

tests/all.rs
liushuyu: Addressed in 7b5fec1f91e961b2fafe6a897b36fd9d06366978

src/builder.rs
liushuyu: > This unfortunately has a side effect where on platforms like Unix this ends up doing a syscall whereas previously accessing `file_type` generally doesn't require a syscall on most major platforms. The intention was that this loop should primarily only do `fs::read_dir` as a sycall and otherwise other syscalls can be avoided.
> 

I have reverted the changes made to this loop.

> My thinking about refactoring this was largely that this loop has one iteration of "look at the file type and delegate appropriately" but above in `append_path_with_name` it's basically the same loop. With the `append_path_with_name` addition it's getting more complicated so I was wondering if it would be possible to refactor this loop and `append_path_with_name` above to call a common function which is the one location where special files are handled.

I am not quite sure which one is better:

1. Make a new delegate function `append_special` and rename the current `append_special` to `append_special_unix`. Then in the `append_special` function, use `fs::metadata` to get the metadata and pass it to `append_special_unix`. In which case, the loop needs to be modified to differentiate a regular file from a special one.
2. Similar to (1) but instead of using `fs::metadata` in that function, `fs::metadata` is used in the "special file" branch of the loop. The advantage of this is one less `stat` when called from `append_path_with_name` (since there is one `stat` in that function as well).
3. Similar to (1) but making `append_special` to accept `Option<fs::metadata>` and `stat` on-demand.

liushuyu:
Ping? Is there update for this Pull Request?

alexcrichton:
Sorry I think this got lost in my inbox, but this looks good to me now!

PR #248, opened 2021-05-18
entry: convert hard link absolute paths to relative ones

gdesmott:
When unpacking a docker image, some hard links may be pointing to, say,
/bin/gunzip. Those were failing the validate_inside_dst() check,
preventing to unpack the image.

Fix this by re-using the same logic as we are already using for actual
files, removing the leading '/' from the link target path.

We do this only if the absolute link is not an ancestor of the target
dir as we want to support this use case according to the existing
absolute_hardlink() test.


src/entry.rs
```diff
@@ -475,6 +475,23 @@ impl<'a> EntryFields<'a> {
                     // links though they're canonicalized to their existing path
                     // so we need to validate at this time.
                     Some(ref p) => {
+                        let src = if src.is_absolute() {
+                            // link points to an absolute path, assume it's relative to the target dir
+                            // by removing the leading '/' as we already do for files, see unpack_in() comments.
```
alexcrichton, line 480: Could this update the commnt above? It seems like that doc block should probably be documenting what's going on here.

 Could this perhaps be structured also a bit differently to do `src.to_path_buf()` early on and only mutate that binding if the condition to trim it is true?

tests/entry.rs
```diff
@@ -62,12 +62,22 @@ fn absolute_hardlink() {
...
     t!(ar.append(&header, &[][..]));
 
+    let mut header = tar::Header::new_gnu();
+    header.set_size(0);
+    header.set_entry_type(tar::EntryType::Link);
+    t!(header.set_path("baz"));
+    // This absolute path on root will be converted when unpacking
+    t!(header.set_link_name("/foo"));
+    header.set_cksum();
+    t!(ar.append(&header, &[][..]));
```
alexcrichton, line 72: There's a couple other branches through the `if` statement added above, could tests be added for the other branches as well?

src/entry.rs
gdesmott: Sure, what about this?

tests/entry.rs
gdesmott: Actually the existing test (`foo`) is already testing the other case:  absolute path starting with the target directory.

alexcrichton:
Looks good to me! I think tests may be failing now though :(

gdesmott:
All tests are passing here (Linux) and I don't have access to a macos system to debug this I'm afraid. :\

gdesmott:
I got access to a macos box from a colleague. Here is the problem:

```
[src/entry.rs:485] &dest_canon = "/private/var/folders/v3/fb_smfzj3bn41k7jj7vfqqh80000gr/T/tarampqNh"
[src/entry.rs:488] src.components() = Components(
    [
        RootDir,
        Normal(
            "var",
        ),
        Normal(
            "folders",
        ),
        Normal(
            "v3",
        ),
        Normal(
            "fb_smfzj3bn41k7jj7vfqqh80000gr",
        ),
        Normal(
            "T",
        ),
        Normal(
            "tarampqNh",
        ),
        Normal(
            "foo",
        ),
    ],
)
[src/entry.rs:492] &src = "var/folders/v3/fb_smfzj3bn41k7jj7vfqqh80000gr/T/tarampqNh/foo"
```

As you can see `Path::components` is parsing `/private` instead of `/` as root. Is that normal or a bug in the parsing? (I have 0 experience with macos)

alexcrichton:
Sorry I'm not sure, I've never seen this behavior before.

gdesmott:
Sorry I got confused while debugging. The actual issue is test's `Tempdir` being created as `/var/folders/...` which is canonicalized to `/private/var/folders/...` as `/var` is a symlink to `/private/var` on macos.
The unpack dest dir was canonicalized but the not the target dir, hence the test failing.

I solved it by canonicalizing the tempdir as well.

alexcrichton:
I think windows may now be failing?

gdesmott:
That's really strange, the test is passing on my Windows box so I'm afraid there is not much more I can do. :\

PR #249, opened 2021-05-20
Extract directory entries after all others when using Archive::unpack.

### LINKED ISSUE #242: Permission error unpacking a directory hierarchy with write-protected folders
When unpacking an archive looking like this:

```
$ tar tvf test.tar.gz
dr-xr-xr-x abbe/staff        0 2021-01-30 17:48 read/
dr-xr-xr-x abbe/staff        0 2021-01-30 17:48 read/only/
-r--r--r-- abbe/staff        0 2021-01-30 17:48 read/only/dirs-above-me
```

using this code 

```rust
use std::fs::File;
use flate2::read::GzDecoder;
use tar::Archive;

fn main() -> Result<(), std::io::Error> {
    let path = "test.tar.gz";

    let tar_gz = File::open(path)?;
    let tar = GzDecoder::new(tar_gz);
    let mut archive = Archive::new(tar);
    archive.unpack(".")?;

    Ok(())
}
```

the result is:

```
Error: Custom { kind: PermissionDenied, error: TarError { desc: "failed to unpack `/Users/abbe/code/tar-permissions/read/only`", io: Custom { kind: PermissionDenied, error: "Permission denied (os error 13) when creating dir /Users/abbe/code/tar-permissions/read/only" } } }
```

This archive works fine to unpack with the `tar` command. This seems to come from the fact that permissions from the tar archive is applied to each folder as it is created and not as an "atomic" operation after unpack [here](https://github.com/alexcrichton/tar-rs/blob/6fa079849506b5044206ab65640fb0f940b69884/src/entry.rs#L443).

afranchuk:
If they are extracted in order, they may fail if a directory's
permissions prevent descendant files from being written. So extracting
them after all other entries ensures that any files that need to be
created can be without being subject to the final permissions.

Closes #242.

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -172,10 +172,22 @@ impl<'a> Archive<dyn Read + 'a> {
         // NotFound exception.
         let dst = &dst.canonicalize().unwrap_or(dst.to_path_buf());
 
+        // Delay any directory entries until the end (they will be created if needed by
+        // descendants), to ensure that directory permissions do not interfer with descendant
+        // extraction.
+        let mut directories = Vec::new();
         for entry in self._entries()? {
             let mut file = entry.map_err(|e| TarError::new("failed to iterate over archive", e))?;
-            file.unpack_in(dst)?;
+            if file.header().entry_type() == crate::EntryType::Directory {
+                directories.push(file);
+            } else {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Seems reasonable to me, thanks!

PR #250, opened 2021-05-26
Avoid zero-mtime files

alexcrichton:
With rust-lang/cargo#9512 discovered this is, uh, odd fallout of the
"deterministic" header mode in this crate. This is an unintended
consequence of the deterministic header mode which should ideally be
fixable. This commit changes the `tar` crate to use a different constant
than 0 when creating archives with the deterministic header mode
(specifically 123456789: Nov 29, 1973). It also will now refuse to
create files with a 0 mtime, instead resetting the mtime to 1 which
should help the mtime be nonzero so tools like lldb don't accidentally
think it's zero.

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -610,6 +610,13 @@ impl<'a> EntryFields<'a> {
 
         if self.preserve_mtime {
             if let Ok(mtime) = self.header.mtime() {
+                // For some more information on this see the comments in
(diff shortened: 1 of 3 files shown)


ehuss (approved):
Seems good to me. Until we run into a tool that is confused by an mtime of 1 😉 .

PR #254, opened 2021-07-21
tidy 'header' implementation

danieleades:
this pull request tidies a couple of tiny nits and aligns the implementation more closely with `async-tar`

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -613,12 +613,9 @@ impl Header {
     /// major device number.
     pub fn set_device_major(&mut self, major: u32) -> io::Result<()> {
         if let Some(ustar) = self.as_ustar_mut() {
-            return {
-                ustar.set_device_major(major);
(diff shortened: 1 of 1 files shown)


src/header.rs
```diff
@@ -615,10 +613,14 @@ impl Header {
     /// major device number.
     pub fn set_device_major(&mut self, major: u32) -> io::Result<()> {
         if let Some(ustar) = self.as_ustar_mut() {
-            return Ok(ustar.set_device_major(major));
+            return {
+                ustar.set_device_major(major);
+                Ok(())
+            };
```
alexcrichton, line 619: I don't think this change is necessary?

alexcrichton: Or otherwise returning a block feels overly complicated, could this be factored as two statements?

danieleades: done.

PR #256, opened 2021-07-23
use field init shorthand

danieleades:


.rustfmt.toml
```diff
@@ -0,0 +1 @@
+use_field_init_shorthand = true
```
alexcrichton, line 1: Can this be deleted? I don't think this needs to be long-lived in this crate.

danieleades: It absolutely can be deleted. However this config would prevent ever having to have this conversation again...

PR #257, opened 2021-07-23
update config and CI with msrv and 'actions-rs'

danieleades:
a few changes here, which will likely require some discussion (hence why i've left this as a draft)

the MSRV i've chosen is arbitrary. I wonder whether testing the MSRV should also be accompanied by a statement in the README. I don't *love* that the MSRV is defined in two places (the clippy config *and* the CI workflow) though it's not too much cognitive load to remember to bump both together.


alexcrichton:
Thanks for the PR, but I'm not looking to make changes like this to CI at this time.

PR #259, opened 2021-08-09
Don't create directories before we've verified they're inside the tar…

benesch:
@alexcrichton @rocallahan this a resubmission of #258 with the rustfmt error fixed. (Sorry, we're a bit impatient to get a new release of tar with this fixed!)


alexcrichton:
Thanks!

PR #262, opened 2021-08-22
Use rust birthday for HeaderMode::Deterministic timestamp

jamessan:
As described in rust-lang/crates.io#3859, the current arbitrary timestamp of 123456789 leads to Debian packages being auto-rejected due to the old timestamp.  This happens for any timestamps before 1975.

Instead of 123456789, use another, more recent arbitrary timestamp -- the timestamp of the first commit for what would become Rust.

https://github.com/graydon/rust-prehistory/commit/b0fd440798ab3cfb05c60a1a1bd2894e1618479e

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -741,8 +741,9 @@ impl Header {
                 //
                 // We just need things to be deterministic here so just pick
                 // something that isn't zero. This time, chosen after careful
-                // deliberation, corresponds to Nov 29, 1973.
-                self.set_mtime(123456789);
+                // deliberation, corresponds to Jul 23, 2006 -- the date of the
+                // first commit for what would become Rust.
+                self.set_mtime(1153704088);
 
                 self.set_uid(0);
                 self.set_gid(0);


Turbo87:
https://github.com/rust-lang/crates.io/issues/3859 talks about 1970-01-01 though, not 1973-11-29. are you sure that this is related?

jamessan:
Yes, I am.  The referenced [Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=976369) is about the timestamps being "too old".  They're currently failing this test because they're set to the epoch.

However, the [specific check](https://salsa.debian.org/lintian/lintian/-/blob/b4a66c3c184bfb6c0f43182b1a638cef81a12608/lib/Lintian/Check/Files/Date.pm#L35-52) that triggers the diagnostic is for anything with an mtime <= the year 1975.

alexcrichton:
Hm well it's a bit frustrating that it's so hard to pick something arbitrary here, but this isn't really any less arbitrary than before...

JohnTitor:
@jamessan Coming from https://github.com/rust-lang/crates.io/issues/3859, so should we enforce cargo to use 0.4.38 to fix the issue reported on crates.io, like https://github.com/rust-lang/cargo/pull/9517?

jamessan:
> Coming from [rust-lang/crates.io#3859](https://github.com/rust-lang/crates.io/issues/3859), so should we enforce cargo to use 0.4.38 to fix the issue reported on crates.io, like [rust-lang/cargo#9517](https://github.com/rust-lang/cargo/pull/9517)?

Yes, please.  I had stopped tracking whether tar-rs had released with this change.  Thank you!

Issue #218, opened 2019-12-09
Support for `Seek` trait


roman-holovin:
As far as I understand, `Archive::skip` is using `read` to skip over file contents. 
Is it possible to require `Read + Seek` and use `seek` to move through the input?

alexcrichton:
Thanks for the report! I originally had `Read` and `Seek` but preferred to only require `Read`. Ideally we'd use some form of specialization for using `seek` where we can, but unfortunately that's not easily done right now.

roman-holovin:
>I originally had Read and Seek but preferred to only require `Read`

Can you describe reasoning or link the discussion? I would like to read that.

>unfortunately that's not easily done right now.

Is there missing feature in compiler to do so? I've tried to modify slightly to require both `Read` and `Seek`, but it seems that is not something `rustc` can do now.

Can it be done with the separate API for now, while the common one is not available? And deprecate it latter.

fermeise:
In our project we have a use case where we need to read the metadata of a tar archive with large file entries (>10 GB). Adding support for `Seek` would allow us to reduce the runtime from minutes down to milliseconds.

I've added a PR with a proposal that adds support for readers that implement `Seek` while maintaining compatibility with readers that don't. Would you consider a change like this? Any alternative proposals?

PR #266, opened 2021-10-04
Add support for readers that implement Seek (#218)

fermeise:
`Archive::new` requires only `Read` for backward-compatibility, while
`Archive::new_from_seek` can be used with readers that also implement
`Seek`, to allow more efficient skipping over file contents.


src/archive.rs
```diff
@@ -143,7 +143,15 @@ impl<R: Read> Archive<R> {
     }
 }
 
-impl<'a> Archive<dyn Read + 'a> {
+impl<R: Read + Seek> Archive<R> {
+    /// Create a new archive using a reader that also implements `Seek`, to
+    /// allow more efficient skipping over file contents.
+    pub fn new_from_seek(obj: R) -> Archive<SeekingSkipRead<R>> {
```
alexcrichton, line 149: I would ideally prefer to avoid exposing `SkipRead` from this crate or the `SeekingSkipRead` type, instead making `Archive<R>` simply "work" if `R` implements `Seek`. That I don't think is natively possible without specialization, but I still think it may be possible to get somewhat close.

I'm imagining something where you'd still use `Archive::new` with seek-able readers, but insteadd of calling `entries` you'd call something like `entries_with_seek` which requires `R: Read + Seek`. The internal implementation details would still be the same as you have today with a custom trait, but the custom trait would be an implementation detail rather than a public API detail.

Would you be up for seeing if this would work?

alexcrichton (commented):
Thanks for this! This is definitely something I think that's worth implementing. I've got a few comments below, but I think this is going in a good direction.

src/skip.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+impl<R: Read + Seek> SkipRead for SeekingSkipRead<R> {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        self.inner.read(buf)
+    }
+
+    fn skip(&mut self, amount: u64) -> io::Result<u64> {
+        let old_pos = self.inner.stream_position()?;
```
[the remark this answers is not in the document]
fermeise: This was motivated by 
> A seek beyond the end of a stream is allowed, but behavior is defined by the implementation.

in https://doc.rust-lang.org/stable/std/io/trait.Seek.html#tymethod.seek
Though I'm not sure if that implies that checking the stream position afterwards is necessary.

src/archive.rs
fermeise: I really like that idea. I uploaded a new version that provides a `entries_with_seek` as you suggested and uses a different mechanism to avoid modifying the public API.

src/archive.rs
```diff
@@ -205,6 +231,14 @@ impl<'a> Archive<dyn Read + 'a> {
     }
 }
 
+impl<'a> Archive<dyn SeekRead + 'a> {
+    fn skip_with_seek(&self, amt: u64) -> io::Result<()> {
+        (&self.inner)
+            .seek(io::SeekFrom::Current(amt as i64))
```
alexcrichton, line 237: Could this use a fallible `i64::try_from(amt)` and return the error as an `io::Error` to catch out-of-bounds cases?

src/archive.rs
```diff
@@ -241,7 +275,11 @@ impl<'a> EntriesFields<'a> {
         loop {
             // Seek to the start of the next header in the archive
             let delta = self.next - self.archive.inner.pos.get();
-            self.archive.skip(delta)?;
+            if let Some(seekable_archive) = self.seekable_archive {
+                seekable_archive.skip_with_seek(delta)?;
```
alexcrichton, line 279: Coudld the `skip_with_seek` method get folded into a helper method of `Entries` to internally dispatch to a read-the-bytes or call-the-`seek` method? I don't think that `Archive::skip` needs to be called anywhere else and I think it'd be good to centralize skipping logic.

alexcrichton (commented):
Looks great to me, thanks!

src/archive.rs
```diff
@@ -153,13 +179,14 @@ impl<'a> Archive<dyn Read + 'a> {
...
             archive: self,
+            seekable_archive,
             done: false,
             next: 0,
             raw: false,
         })
     }
 
-    fn _unpack(&mut self, dst: &Path) -> io::Result<()> {
+    fn _unpack(&'a mut self, dst: &Path) -> io::Result<()> {
```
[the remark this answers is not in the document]
fermeise: rustc seems to need this to determine the lifetime of the `seekable_archive` argument when calling `_entries()`.

src/archive.rs
```diff
@@ -205,6 +232,16 @@ impl<'a> Archive<dyn Read + 'a> {
...
 
+impl<'a> Archive<dyn SeekRead + 'a> {
+    fn skip_with_seek(&self, amt: u64) -> io::Result<()> {
+        (&self.inner)
+            .seek(io::SeekFrom::Current(
+                i64::try_from(amt).map_err(|_| other("seek position out of bounds"))?,
+            ))
+            .map(|_| ())
+    }
+}
```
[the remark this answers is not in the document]
fermeise: `pos` was updated in `ArchiveInner::seek`, which is now folded. I added some tests for reading and skipping file contents.

PR #270, opened 2021-11-09
Added support for target wasm32-wasi

john-sharratt:
This allows the tar to work in a browser, wasmer and WASI platforms.

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -153,7 +153,7 @@ impl<R: Read> Archive<R> {
 fn default_preserve_mtime() -> bool {
     if cfg!(target_os = "wasi") {
         // The filetime crate does not yet implement updating file times but hopefully
-        // the upstream developers will incorporate it at which point at which time we
+        // the upstream developers will incorporate it at which time this we
         // can default again to preserving mtime on WASI systems
         // see: https://github.com/alexcrichton/filetime/blob/master/src/wasm.rs#L34
         return false;
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -13,6 +13,7 @@ use crate::error::TarError;
 use crate::other;
 use crate::pax::pax_extensions_size;
 use crate::{Entry, GnuExtSparseHeader, GnuSparseHeader, Header};
+use crate::header::canonicalize;
 
 /// A top-level representation of an archive file.
 ///
@@ -198,10 +199,7 @@ impl Archive<dyn Read + '_> {
         // extended-length path with a 32,767 character limit. Otherwise all
         // unpacked paths over 260 characters will fail on creation with a
         // NotFound exception.
-        #[cfg(not(target_os = "wasi"))]
-        let dst = &dst.canonicalize().unwrap_or(dst.to_path_buf());
-        #[cfg(target_os = "wasi")]
-        let dst = &dst.to_path_buf();
+        let dst = &canonicalize(&dst);
 
         // Delay any directory entries until the end (they will be created if needed by
         // descendants), to ensure that directory permissions do not interfer with descendant
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -7,6 +7,7 @@ use std::io::{self, SeekFrom};
 use std::marker;
 use std::path::Path;
 
+#[allow(unused_imports)]
 use crate::entry::{EntryFields, EntryIo};
 use crate::error::TarError;
 use crate::other;
@@ -197,7 +198,10 @@ impl Archive<dyn Read + '_> {
         // extended-length path with a 32,767 character limit. Otherwise all
         // unpacked paths over 260 characters will fail on creation with a
         // NotFound exception.
+        #[cfg(not(target_os = "wasi"))]
         let dst = &dst.canonicalize().unwrap_or(dst.to_path_buf());
+        #[cfg(target_os = "wasi")]
+        let dst = &dst.to_path_buf();
 
         // Delay any directory entries until the end (they will be created if needed by
         // descendants), to ensure that directory permissions do not interfer with descendant
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -56,7 +56,7 @@ impl<R: Read> Archive<R> {
             inner: ArchiveInner {
                 unpack_xattrs: false,
                 preserve_permissions: false,
-                preserve_mtime: true,
+                preserve_mtime: default_preserve_mtime(),
                 overwrite: true,
                 ignore_zeros: false,
                 obj: RefCell::new(obj),
@@ -148,6 +148,17 @@ impl<R: Read> Archive<R> {
     pub fn set_ignore_zeros(&mut self, ignore_zeros: bool) {
         self.inner.ignore_zeros = ignore_zeros;
     }
+}
+
+fn default_preserve_mtime() -> bool {
+    if cfg!(target_os = "wasi") {
+        // The filetime crate does not yet implement updating file times but hopefully
+        // the upstream developers will incorporate it at which point at which time we
+        // can default again to preserving mtime on WASI systems
+        // see: https://github.com/alexcrichton/filetime/blob/master/src/wasm.rs#L34
+        return false;
+    }
+    true
 }
 
 impl<R: Seek + Read> Archive<R> {
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -10,10 +10,10 @@ use std::path::Path;
 #[allow(unused_imports)]
 use crate::entry::{EntryFields, EntryIo};
 use crate::error::TarError;
+use crate::header::canonicalize;
 use crate::other;
 use crate::pax::pax_extensions_size;
 use crate::{Entry, GnuExtSparseHeader, GnuSparseHeader, Header};
-use crate::header::canonicalize;
 
 /// A top-level representation of an archive file.
 ///
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -10,10 +10,10 @@ use std::path::{Component, Path, PathBuf};
 use crate::archive::ArchiveInner;
 use crate::error::TarError;
 use crate::header::bytes2path;
+use crate::header::canonicalize_or_err;
 use crate::other;
 use crate::pax::pax_extensions;
 use crate::{Archive, Header, PaxExtensions};
-use crate::header::canonicalize_or_err;
 
 /// A read-only view into an entry of an archive.
 ///
@@ -617,7 +617,7 @@ impl<'a> EntryFields<'a> {
 
         if self.preserve_mtime {
             if let Ok(mtime) = self.header.mtime() {
-                set_mtime(dst, &mut f, mtime)?;                
+                set_mtime(dst, &mut f, mtime)?;
             }
         }
         if let Ok(mode) = self.header.mode() {
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -722,7 +722,7 @@ impl<'a> EntryFields<'a> {
             preserve: bool,
         ) -> io::Result<()> {
             if cfg!(target_os = "wasi") {
-                // WASI file permissions have not yet been implemented in the standard library panics
+                // WASI file permissions have not yet been implemented in the standard library
                 // see: https://github.com/rust-lang/rust/blob/master/library/std/src/sys/wasi/fs.rs#L538
                 if preserve {
                     Err(io::Error::new(io::ErrorKind::Other, "setting file permissions not yet supported on WASI - omit preserve_permissions to avoid this error"))
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -7,9 +7,6 @@ use std::io::{self, Error, ErrorKind, SeekFrom};
 use std::marker;
 use std::path::{Component, Path, PathBuf};
 
-#[allow(unused_imports)]
-use filetime::{self, FileTime};
-
 use crate::archive::ArchiveInner;
 use crate::error::TarError;
 use crate::header::bytes2path;
@@ -574,7 +571,6 @@ impl<'a> EntryFields<'a> {
         fn open(dst: &Path) -> io::Result<std::fs::File> {
             OpenOptions::new().write(true).create_new(true).open(dst)
         }
-        #[allow(unused_mut)]
         let mut f = (|| -> io::Result<std::fs::File> {
             let mut f = open(dst).or_else(|err| {
                 if err.kind() != ErrorKind::AlreadyExists {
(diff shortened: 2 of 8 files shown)


alexcrichton:
Thanks for the PR here, but I would prefer to not add support here by sprinkling a lot of `#[cfg]` throughout the code. I think it's fine for wasm32-wasi to behave a bit differently for example by having different defaults for unpacking and similar. For example not preserving mtime and avoiding xattrs seem like reasonable defaults for wasi, but otherwise I don't think it's necessary to `#[cfg]` away the code. Can you speak more to why `canonicalize` is necessary to specialize? 

Would it be possible to add tests for this as well?

john-sharratt:
canonicalize is not implemented in the standard library for wasi so it panics [cry]
https://github.com/rust-lang/rust/blob/master/library/std/src/sys/wasi/fs.rs#L609

Yeah I don't like the #[cfg] macros either - I can try and remove some but its not easy to be honest - alternative is to fork tar-rs until WASI gets more support - or maybe you have another idea - still not cracked this problem myself.

Testing WASI is not very easy at the moment as its quite hard to get unit tests to run without panicing - the good thing is that the basics of the toolchain are there already and its more the lack of support in the standard library thats holding it back, basically you can test by running:

```sh
cargo wasi test
```

But it fails with this:
```
Caused by:
    0: failed to invoke command default
    1: wasm trap: unreachable
       wasm backtrace:
           0: 0x178c2d - <unknown>!__rust_start_panic
           1: 0x13e951 - <unknown>!rust_panic
           2: 0xcadf1 - <unknown>!std::panicking::rust_panic_with_hook::h308937c11f14a0b0
           3: 0x17647b - <unknown>!std::panicking::begin_panic::{{closure}}::hebc9643fe2b695d7
           4: 0x1753f7 - <unknown>!std::sys_common::backtrace::__rust_end_short_backtrace::he2536be214bafdb9
           5: 0x17644e - <unknown>!std::panicking::begin_panic::hd2671e09b3b4e5c3
           6: 0x178762 - <unknown>!std::sys::wasi::os::temp_dir::hbc626cdbc609eb38
           7: 0x178b02 - <unknown>!std::env::temp_dir::h25ed3bd578de46c3
           8: 0x166ba2 - <unknown>!tempfile::Builder::tempdir::h702dce260a633d0d
           9: 0x60aa5 - <unknown>!all::append_dir_all_blank_dest::h888625bbe2adcd99
          10: 0x17180e - <unknown>!all::append_dir_all_blank_dest::{{closure}}::h2f896a4f4c3d3da9
          11: 0x1735cd - <unknown>!core::ops::function::FnOnce::call_once::h9d99a1526320ffc5
          12: 0x176de4 - <unknown>!test::__rust_begin_short_backtrace::h44c13f5e334be418
          13: 0x178a5d - <unknown>!core::ops::function::FnOnce::call_once{{vtable.shim}}::h99a922c74aed367b
          14: 0x92f6b - <unknown>!test::run_test::run_test_inner::h60a67529b5c5db98
          15: 0x88c1a - <unknown>!test::run_test::h2251c1b970756ec5
          16: 0x2c99e - <unknown>!test::console::run_tests_console::h3dae2ea08348f81f
          17: 0x8e102 - <unknown>!test::test_main::h29ec7696c5542e21
          18: 0x12da49 - <unknown>!test::test_main_static::hcf5afa2aa7e5d486
          19: 0x177f7f - <unknown>!all::main::hd66e46cbdbfb7162
          20: 0x17121a - <unknown>!core::ops::function::FnOnce::call_once::hf6aa8f732d7f16b7
          21: 0x17054f - <unknown>!std::sys_common::backtrace::__rust_begin_short_backtrace::habdcb39f4c03c5cb
          22: 0x16ca05 - <unknown>!std::rt::lang_start::{{closure}}::h61f52bbcf039bda5
          23: 0x12e68d - <unknown>!std::rt::lang_start_internal::hcfff2f9346b7fb75
          24: 0x15c6b3 - <unknown>!std::rt::lang_start::hf92037ed41a12602
          25: 0x177ef0 - <unknown>!__original_main
          26: 0x178407 - <unknown>!_start
          27: 0x178ae9 - <unknown>!_start.command_export
       note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable to may show more debugging information
       
error: failed to execute "wasmtime" "--" "/prog/tar-rs/target/wasm32-wasi/debug/deps/all-c4ccc64a95c19939.wasm"
    status: exit code: 13
```

Reason is the tempfs is not implemented in the std library either - this I've worked around in other projects and hopefully the rust std library developers will patch this soon and the unit tests will come back online.

For now the best we can do without manual testing or rewriting the unit tests (which will be for nothing when they fix up the last bits of the std library) is to run a basic compile test
```sh
cargo wasi build
```

alexcrichton:
If you'd prefer to fork that's ok by me, but if you'd like to contribute to this upstream I'd like items in review to be taken care of (such as cutting down on `#[cfg]`). There's various ways to do that, such as having one helper method that's defined differnetly for wasi/else, instead of having a `#[cfg]`-per-use.

For tests I think it's ok to not have the whole test suite working, but presumably you care about something working and if you want that to keep working over time I'd recommend adding tests to CI.

john-sharratt:
Don't get me wrong - I think its always better to work and contribute on the upstream and my comment was only a request if you really didn't like WASI due to its "new-ness" then I understand.

Let me take a look at the #[cfg]'s and see what I can do to reduce them.

For the test suite the only way I can get it to work is if I disable a chunk of them for WASI and try to get one or two of them working but I don't know a way to do that without the #[cfg] macros.

Totally open to suggestions

Cheers

john-sharratt:
So I did another commit and addressed the concerns you raised.

Basically I reduced the number of #[cfg] lines dropped in and made it more cleaner by splitting out the canonicalize function and mocking another implementation of it for WASI.

Second I worked through the list of unit tests and got 28 of the total working with the WASI test so while it does not give good coverage on the file system level (given WASI isn't there yet anyway) but the rest of the tar functionality has some decent coverage. To integrate with your CI/CD you need to run the "cargo wasi test" command... you should get this:

```
cargo wasi test -- --nocapture
   Compiling tar v0.4.37 (/prog/tar-rs)
    Finished test [unoptimized + debuginfo] target(s) in 0.70s
     Running unittests (target/wasm32-wasi/debug/deps/tar-6fd696d6c645a524.wasm)
     Running tests/all.rs (target/wasm32-wasi/debug/deps/all-d49bb81736cacecb.wasm)
     Running tests/entry.rs (target/wasm32-wasi/debug/deps/entry-1fc4c15213d03354.wasm)
     Running `/prog/tar-rs/target/wasm32-wasi/debug/deps/tar-6fd696d6c645a524.wasm --nocapture`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running `/prog/tar-rs/target/wasm32-wasi/debug/deps/all-d49bb81736cacecb.wasm --nocapture`

running 28 tests
test append_long_multibyte ... ok
test header::byte_slice_conversion ... ok
test header::default_gnu ... ok
test header::dev_major_minor ... ok
test header::extended_numeric_format ... ok
test header::goto_old ... ok
test header::goto_ustar ... ok
test header::mtime ... ok
test header::set_ustar_path_hard ... ok
test header::user_and_group_name ... ok
test header_impls ... ok
test header_impls_missing_last_header ... ok
test header_size_overflow ... ok
test links ... ok
test long_linkname_trailing_nul ... ok
test long_name_trailing_nul ... ok
test octal_spaces ... ok
test pax_linkpath ... ok
test pax_path ... ok
test pax_simple ... ok
test pax_size ... ok
test reading_entries ... ok
test reading_entries_with_seek ... ok
test reading_files ... ok
test reading_sparse ... ok
test simple_concat ... ok
test skipping_entries_with_seek ... ok
test sparse_with_trailing ... ok

test result: ok. 28 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running `/prog/tar-rs/target/wasm32-wasi/debug/deps/entry-1fc4c15213d03354.wasm --nocapture`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

src/archive.rs
```diff
@@ -7,11 +7,13 @@ use std::io::{self, SeekFrom};
 use std::marker;
 use std::path::Path;
 
+#[allow(unused_imports)]
```
alexcrichton, line 10: Is this attribute still necessary? If so I think that w/e is unused on wasi should probably get moved to w/e `#[cfg]`-specific code its used within.

src/entry.rs
```diff
@@ -616,6 +619,7 @@ impl<'a> EntryFields<'a> {
             )
         })?;
 
+        #[cfg(not(target_os = "wasi"))]
```
alexcrichton, line 622: Is this necessary? I think it's reasonable that if `preserve_mtime` is requested that it generates an error on wasi since it's not supported. If `preserve_mtime` is on-by-default then it can be off-by-default for wasi.

src/entry.rs
```diff
@@ -631,9 +635,11 @@ impl<'a> EntryFields<'a> {
                 })?;
             }
         }
+        #[cfg(not(target_os = "wasi"))]
```
alexcrichton, line 638: I think this would make more sense as a conditional definition of `set_perms` for wasi (which I think already exists otherwise for windows/unix)

src/entry.rs
```diff
@@ -631,9 +635,11 @@ impl<'a> EntryFields<'a> {
...
         if let Ok(mode) = self.header.mode() {
             set_perms(dst, Some(&mut f), mode, self.preserve_permissions)?;
         }
+        #[cfg(not(target_os = "wasi"))]
```
alexcrichton, line 642: Like above I don't think a `#[cfg]` here is necessary, I think it suffices if the default for` unpack_xattrs` is `false` for wasi.

src/header.rs
```diff
@@ -1614,15 +1614,34 @@ pub fn bytes2path(bytes: Cow<[u8]>) -> io::Result<Cow<Path>> {
...
 #[cfg(target_arch = "wasm32")]
 fn invalid_utf8<T>(_: T) -> io::Error {
     io::Error::new(io::ErrorKind::InvalidData, "Invalid utf-8")
 }
+
+pub fn canonicalize(path: &Path) -> PathBuf {
+    canonicalize_or_err(path).unwrap_or(path.to_path_buf())
+}
+
+#[cfg(not(target_os = "wasi"))]
```
alexcrichton, line 1634: I don't think that `#[cfg]` is necessary here, I think it's fine to use `if cfg!(...)` and can you also add a comment for why this is specialized on wasi?

tests/all.rs
```diff
@@ -130,6 +131,7 @@ fn reading_files() {
 }
 
 #[test]
+#[cfg(not(target_os = "wasi"))]
```
alexcrichton, line 134: Could this be done as:

```
#[cfg_attr(target_os = "wasi", ignore)]
```

instead?

alexcrichton (commented):
Thanks! Can CI also be updated with compiling to wasi and running tests?

src/archive.rs
john-sharratt: It was only used in one line of code so I removed the import instead and just fully qualified the function

src/entry.rs
john-sharratt: I moved this logic into the function as you suggested and put some comments around why it is not supported.
also I implemented a default_preserve_mtime() function which turns it off by default

src/entry.rs
john-sharratt: This will now return an error if someone uses this library with WASI and turns the preserve_permissions flag on.
Note if someone changes the default in the future this will now cause many builds to fail but changing the default on this flag would be a breaking change anyway so this is kind of already a risk

src/header.rs
john-sharratt: Refactored the code to use if cfg!(...) instead as you suggested

src/entry.rs
john-sharratt: In the end this actually wasnt needed at all given the default implementation on WASM is just to return Ok(()). I removed this attribute

tests/all.rs
john-sharratt: I've refactored the code to change to ignore as you suggested - looks good after my testing

john-sharratt:
> Thanks! Can CI also be updated with compiling to wasi and running tests?

Sure - its a simple change for you to implement

Just use this for builds...
```sh
cargo wasi build
```

...and use this for tests....
```sh
cargo wasi test -- --nocapture
```

The tool-chain can be installed with this command...
```sh
cargo install cargo-wasi
```

john-sharratt:
There we go... added some lovely WASI unit tests in the CI/CD
When you merge this will you also publish? There are some downstream projects depending on this.

src/entry.rs
```diff
@@ -639,6 +628,31 @@ impl<'a> EntryFields<'a> {
...
+        #[cfg_attr(target_os = "wasi", allow(unused_variables))]
+        fn set_mtime(dst: &Path, f: &mut std::fs::File, mtime: u64) -> Result<(), TarError> {
+            if cfg!(target_os = "wasi") {
+                // The filetime crate does not yet implement updating file times but hopefully
+                // the upstream developers will incorporate it at which time this can be
+                // implemented in this crate as well.
+                // see: https://github.com/alexcrichton/filetime/blob/master/src/wasm.rs#L34
+                return Err(TarError::new(format!("failed to set mtime for `{}`", dst.display()),
+                    io::Error::new(io::ErrorKind::Other, "setting modified date/times is not yet supported on WASI - omit preserve_mtime to avoid this error")));
+            }
```
alexcrichton, line 640: Does this need to be specifically handled for wasi? Could the current error be deferred to instead of explicitly returning an error?

src/entry.rs
```diff
@@ -705,9 +719,19 @@ impl<'a> EntryFields<'a> {
             dst: &Path,
             f: Option<&mut std::fs::File>,
             mode: u32,
-            _preserve: bool,
+            preserve: bool,
         ) -> io::Result<()> {
-            Err(io::Error::new(io::ErrorKind::Other, "Not implemented"))
+            if cfg!(target_os = "wasi") {
```
alexcrichton, line 724: Like above I don't think that this needs specific logic for wasi/wasm, if the goal is to get a better error message then I think this can improve the error message in-line and there's no need to do something different for wasi vs other wasm targets.

src/entry.rs
john-sharratt: Sure no problem - checked the code and it will work fine like that so removed this call branch

src/entry.rs
john-sharratt: Unfortunately this is not like the other one above because of this...
https://github.com/rust-lang/rust/blob/master/library/std/src/sys/wasi/fs.rs#L538

Basically without the check with an error message the process will panic rather than return a graceful error - I think the majority of consumers would prefer an error message over a panic.

alexcrichton: I think this would be best handled with a new configuration option similar to preserving mtimes to preserving permissions which is on-by-default but off-by-default for wasi.

src/header.rs
```diff
@@ -1614,15 +1614,34 @@ pub fn bytes2path(bytes: Cow<[u8]>) -> io::Result<Cow<Path>> {
...
 }
+
+pub fn canonicalize(path: &Path) -> PathBuf {
+    canonicalize_or_err(path).unwrap_or(path.to_path_buf())
+}
+
+pub fn canonicalize_or_err(path: &Path) -> io::Result<PathBuf> {
+    if cfg!(target_os = "wasi") {
+        // canonicalize is not yet supported in the standard library
+        // see: https://github.com/rust-lang/rust/blob/master/library/std/src/sys/wasi/fs.rs#L609
```
alexcrichton, line 1637: I think this comment can be refined to how the wasi platform doesn't have a canonicalize method right now.

john-sharratt: Not sure what you mean - I linked to the exact line of code that isn't implemented.
How can I refine that more :-)
Cheers

src/entry.rs
john-sharratt: It already has a preserve flag on line 719 which itself comes from the preserve_permissions which indeed is already following the "off-by-default for wasi" logic which I implemented in the last commit

john-sharratt:
I'm out of investment tokens on this one now... not enough hours in the day unfortunately so will only be able to do minimal updates from here. I am sure others can help you though now that its well commented, tested and you have CI/CD.
Cheers

alexcrichton:
Ok no worries, if someone else wants to come pick this up then this can be reopened.

alexcrichton:
If you're willing to go through the rigamarole of publishing a whole new fork on crates.io I'd ask again if you're willing to address the final few comments on review. If you would prefer to publish a fork I have no preference on the name.

john-sharratt:
I told you I have a preference for merging upstream but I'm not going to iterate over and over on few really minor points that don't make any real difference, my hours in the day are extremely precious to me and I've put a ton of work into this already so it's not like I didn't try.

It's not that a blame you, it's obvious you really love your code to be a very particular way... I'm cool with that... so why don't you just accept the merge and change the lines to be exactly how you want them to meet your particular style of coding? I do it all the time - it's often easier.

Anyhow I'll go publish on cargo tomorrow and if you accept the merge later just let me know and I'll yank it out and kill it... its better to clean up the copied packages for the communities sake. Ultimately though there are downstream WASI projects that also need help that I need to work on urgently.

john-sharratt:
How about I give it one last try tomorrow; we both put effort into this and we should try our best to get it over the line - what are the last changes you would like me to make as prescriptively as possible.

Issue #192, opened 2019-04-19
adding a long hardlink without _fs APIs


the8472:
Similar to #108, the internal helper method for [appending long links](https://github.com/alexcrichton/tar-rs/blob/7473c27372a8adaf946e61836eeabf36ed2e25dc/src/builder.rs#L467-L479) is only accessible through APIs that attempt to access the filesystem.

It would be great if it were possible to pass a header (with symlink or hardlink `entry_type`) and a `name` + `link_name` as `Path` and have the builder insert the necessary long-name headers. I.e. an equivalent to `append_data`

Maybe some general intermediate-level API is missing for adding headers and data separately to orthogonalize the construction of complicated headers that must be written in several parts (long name, long links, sparse header) and their associated data chunks. A single function taking several enums to cover all those combinations might work too.

cgwalters:
I'm hitting this in https://github.com/ostreedev/ostree-rs-ext/issues/162

I'm willing to work on a PR here.  But reading the code, one thing that strikes me as somewhat odd is that we have this "base interface" in `tar::Header` and APIs to "cast" it to e.g. a `GnuHeader`.

But why do we not directly expose all the API methods like `set_link_name()` on `GnuHeader`?  In my case, I know I am writing a GNU-formatted tar archive.  I always want that behavior.

It seems weird to me how many of the `Header` methods do things like:

```
    fn _set_path(&mut self, path: &Path) -> io::Result<()> {
        if let Some(ustar) = self.as_ustar_mut() {
            return ustar.set_path(path);
        }
        ...
    // Or base case here
```

The inverse problem here is that the `append_data()` behavior of trying the base case but silently "upgrading" to GNU may not be what some people want - they may want to fail instead of writing a GNU entry.  I admit this use case is probably obscure, but to rephrase I think we should be consistent in our interface - the format of an archive *as a whole* is either GNU or PAX or base, etc.

cgwalters:
PR in https://github.com/alexcrichton/tar-rs/pull/273

cgwalters:
(Regarding my confusion around `Header` above, the reason we need a special case `append_data()` is internally it's represented as *two* tar entries, so we need to handle it via `Builder`)

PR #273, opened 2021-11-18
Add an `append_link()` method to handle long link targets

cgwalters:
We should support appending long symlink targets, because this
occurs in real world filesystems.  In my case, RPM set up
`.build-id` symlinks which can get long.

Add an `append_link()` method following the precendent of
`append_path()` - we're just supporting *two* potentially
long filenames.

As a side benefit, we can just do the `std::io::empty()` dance
internally and not require the caller to specify it.

The addition of special case `append()` methods is unfortunate,
because the header API methods are then really an attractive nuisance.
We should potentially consider deprecating them.

Closes: #192

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -162,6 +162,54 @@ impl<W: Write> Builder<W> {
         self.append(&header, data)
     }
 
+    /// Adds a new link (symbolic or hard) entry to this archive with the specified path and target.
+    ///
+    /// This function is similar to [`Self::append_data`] which supports long filenames,
+    /// but also supports long link targets using GNU extensions if necessary.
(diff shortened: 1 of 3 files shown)


cgwalters:
This passes my unit test, but I'm still doing some more real world testing and comparing this (and my higher level code) with what GNU `tar` actually does.

cgwalters:
OK this works, I tested in concert with https://github.com/ostreedev/ostree-rs-ext/pull/165
What delayed me a bit was debugging https://github.com/ostreedev/ostree-rs-ext/pull/164 because in this code it appears as object (SHA-256) corruption, and it wasn't immediately obvious to me that the link name being truncated was the source.

alexcrichton:
Thanks!

PR #276, opened 2022-01-06
archive: allow preserving ownerships when unpacking

liushuyu:
Added support for unpacking files while preserving numeric ownership IDs on Unix, this is useful when you want to make a backup utility since ownerships are very important on Unix-like systems. (Also useful if you want to make a system installer, the underlying idea is the same.)

Currently, operations are only available on Unix.
On Windows, it's also possible to set user ownerships, however, it's a very different system and does not use numeric IDs AFAIK. Any other suggestions on the current implementations are welcome!


tests/all.rs
```diff
@@ -1385,3 +1385,33 @@ fn header_size_overflow() {
...
+        assert_eq!(meta.uid(), 580800001);
+        assert_eq!(meta.gid(), 580800000);
+        let meta = std::fs::metadata(td.path().join("iamuid580800002")).unwrap();
+        assert_eq!(meta.uid(), 580800002);
+        assert_eq!(meta.gid(), 580800002);
+    } else {
+        // it's not possible to unpack tar while preserving ownership
+        // without root permissions
+        assert!(ar.unpack(td.path()).is_err());
+    }
```
alexcrichton, line 1416: AFAIK this is unlikely to ever get executed, even in CI, so could a test be written that runs in normal user mode as well?

src/entry.rs
```diff
@@ -635,11 +642,87 @@ impl<'a> EntryFields<'a> {
...
             set_perms(dst, Some(&mut f), mode, self.preserve_permissions)?;
         }
+        if self.preserve_ownerships {
+            set_ownerships(
+                dst,
+                Some(&mut f),
+                self.header.uid().ok(),
+                self.header.gid().ok(),
+            )?;
+        }
```
alexcrichton, line 652: This pattern of "set all the things and metadata" happens in a few places in this function, could they be refactored where one function "sets all the things" and internally it dispatches to ownership, permission, mtime, etc?

liushuyu: Sure.

tests/all.rs
liushuyu: > AFAIK this is unlikely to ever get executed, even in CI, so could a test be written that runs in normal user mode as well?


I think it would be very difficult. `chown` itself is a privileged operation, even on Windows. So it would be not possible to test it properly using normal user permissions.

Although, there might be a way to change `gid` on a file that is currently owned by the current effective user (file `uid` == user `euid`). But this kind of test would not cover the `uid` change, so it would be very difficult if not impossible to thoroughly test this functionality under a normal user account.

src/entry.rs
```diff
@@ -444,13 +445,36 @@ impl<'a> EntryFields<'a> {
...
+        fn set_perms_ownerships(
+            dst: &Path,
+            f: Option<&mut std::fs::File>,
+            header: &Header,
+            perms: bool,
+            ownerships: bool,
+        ) -> io::Result<()> {
+            // ownerships need to be set first to avoid stripping SUID bits in the permissions ...
+            if ownerships {
+                set_ownerships(dst, &f, header.uid().ok(), header.gid().ok())?;
```
alexcrichton, line 457: I think this would want to propagate the error for reading the uid/gid instead of ignoring them perhaps?

src/entry.rs
```diff
@@ -632,14 +660,86 @@ impl<'a> EntryFields<'a> {
...
+        #[cfg(unix)]
+        fn _set_ownerships(
+            dst: &Path,
+            f: &Option<&mut std::fs::File>,
+            uid: Option<u64>,
+            gid: Option<u64>,
+        ) -> io::Result<()> {
+            use std::os::unix::prelude::*;
+
+            // -1 (0xffffffff) means no ownership change in most libc implementations
```
alexcrichton, line 704: This comment about "most libc implementations" is a little worrisome, what about the "other" libc implementations? Does this simply not work in those situations?

(I'm not sure why uid/gid here are optional though)

tests/all.rs
alexcrichton: If that is the situation can you arrange for the test to happen with `sudo` on CI?

Also ideally tthis wouldn't check in a `*.tar` file but build it up here so the tar file can be easily changed as well.

src/entry.rs
liushuyu: According to https://pubs.opengroup.org/onlinepubs/9699919799/functions/chown.html, `-1` is a POSIX-standard way to keep the current ID. There are probably some `libc` implementations that are not fully POSIX-compatible that do not understand this.

But yeah, if we want to handle the errors when reading UID/GID then this is irrelevant.

src/entry.rs
```diff
@@ -632,14 +660,83 @@ impl<'a> EntryFields<'a> {
...
+            f: &Option<&mut std::fs::File>,
+            uid: u64,
+            gid: u64,
+        ) -> io::Result<()> {
+            use std::os::unix::prelude::*;
+
+            match f {
+                Some(f) => unsafe {
+                    let fd = f.as_raw_fd();
+                    if libc::fchown(fd, uid as u32, gid as u32) != 0 {
```
alexcrichton, line 707: Oh sorry one thing I forgot, I think this will want to use `try_into` instead of `as u64` to avoid silently truncating as well.

src/entry.rs
```diff
@@ -632,14 +660,83 @@ impl<'a> EntryFields<'a> {
...
+                    }
+                },
+                None => unsafe {
+                    let path = std::ffi::CString::new(dst.as_os_str().as_bytes()).map_err(|e| {
+                        io::Error::new(
+                            io::ErrorKind::Other,
+                            format!("path contains null character: {:?}", e),
+                        )
+                    })?;
+                    if libc::lchown(path.as_ptr(), uid as u32, gid as u32) != 0 {
```
alexcrichton, line 720: (ditto with above)

alexcrichton (commented):
Thanks! I think that the `ownership.tar` can be removed now as well?

src/entry.rs
liushuyu: Fixed.

PR #277, opened 2022-01-07
Make the dependency filetime optional

chris-morgan:
filetime is in the same boat as xattr: only used if you’re unpacking tarballs to disk, and already disableable at runtime (`preserve_mtime`, `unpack_xattrs`). An excellent candidate for making optional.

This is a very slightly breaking change: anyone using tar with default features disabled (to lose xattr) will now also lose filetime. I’d consider that narrowly acceptable in the circumstances of the similarity of the two optional dependencies.


chris-morgan:
I should have put the code down while I still had a chance.

I’ve pushed two more commits, one making Builder optional in order to futilely seek zero dependencies, and one to unbreak Builder on `cfg(not(any(unix, windows)))` platforms, e.g. wasm32-wasi should work now instead of panicking (untested), and anything else that’s also not wasm32 should work instead of not compiling. (When I flip the unix and not(unix) on fill_platform_from, the tests still pass; not sure how much weight to put on that, though—clearly nothing’s testing fine mode preservation.) Details in commit messages. This branch name is now silly. Such is life.

Up to you what you do with this. If you decide you don’t like splitting out an optional builder feature, I hope you at least consider the fill_platform_from patch (with the very minor changes required to disentangle it).

This was fun, even if you don’t like it all. 🙂

alexcrichton:
Is there any particular reason for making the dependency optional?

chris-morgan:
It’s just because major use cases don’t actually use filetime (e.g. I’m using tar and reading from the tarball without unpacking), so I think it’s nice to remove it from the dependency tree proactively rather than having to compile it and then wait until later in compilation or linking to confirm that it wasn’t actually used. (Mildly related: that’s been a pet peeve of mine in license-scraping tools: they don’t recognise when nothing actually gets used from a crate, like this.) It’s the principle of the thing, keeping things lean deliberately so that the compiler can definitely avoid needing to compile unused dependencies. Otherwise, why is xattrs marked as optional?

glandium:
Also, filetime now pulls in windows-sys, which is a large dependency that is still a moving target, so many crates also depend on different versions of it.

PR #281, opened 2022-03-02
Append entry

### LINKED ISSUE #279: filtering entries before unpacking
I have a tar ball with windows libraries in it with case-sensitive symlinks to files to match the behaviour on a case-insensitive file system. When extracting the tar on a case insensitive file system things blow up. I'm wondering what the best way would be to filter out symlinks on case insensitive file systems while keeping them on case sensitive ones.

dvc94ch:
This revives #178 and fixes #279 and #280


dvc94ch:
will need a little more work, the symlinks need to be buffered/sorted and then written at the end.

dvc94ch:
Closing for now. I think the real problem is that tar is a unix format. For it to be a proper cross platform format it would have to encode with the symlink if it's a file or a directory symlink. Anything else is really just a hack.

Issue #288, opened 2022-04-02
Symlink mtimes aren't preserved


jrwats:
Regular file mtimes are preserved with `preserve_mtime = true`, symlinks should behave similarly.

marcospb19:
Hey @jrwats, I think you forgot to close this issue after it was solved in the PR above.

PR #289, opened 2022-04-02
Preserve symlink mtimes

jrwats:
An archive with `preserve_mtime = true` should match the behavior of `tar -vpxf ...`

Addresses: https://github.com/alexcrichton/tar-rs/issues/288

### DIFF
--- a/tests/all.rs
+++ b/tests/all.rs
@@ -792,6 +792,10 @@ fn unpack_links() {
 
     let md = t!(fs::symlink_metadata(td.path().join("lnk")));
     assert!(md.file_type().is_symlink());
+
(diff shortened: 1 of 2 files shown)


tests/all.rs
```diff
@@ -792,6 +792,10 @@ fn unpack_links() {
 
     let md = t!(fs::symlink_metadata(td.path().join("lnk")));
     assert!(md.file_type().is_symlink());
+
+    let mtime = FileTime::from_last_modification_time(&md);
+    assert_eq!(mtime.unix_seconds(), 1448291033);
```
jrwats, line 797: Previously this would have been "now"

PR #293, opened 2022-06-06
feat: unpack_with_filter(); append_dir_with_filter()

alsuren:
This seems related to https://github.com/alexcrichton/tar-rs/issues/279

in https://github.com/cargo-quick/cargo-quick/issues/14 I find myself wanting to create something similar to Docker layers, using the `tar` crate.

I want to be able to keep track of a list of all files that I unpacked, then do my `cargo build` and then unpack only the files that are new or have changed (I'm assuming that I won't have any deleted files, but if I do, the OCI spec has a way of encoding tombstones in layers, so I can copy that).

In theory, I could take control of the directory walking myself (that's what we do in https://github.com/cargo-quick/kache/blob/main/src/tar.rs) but when I started prototyping cargo-quickbuild, I used `append_dir_all()` and `.unpack()`, 

unresolved questions:
- [ ] there is an asymmetry between the filter for `unpack_with_filter()` and the one for `append_dir_with_filter()`. Should they be reconciled?
- [ ] is it okay that I don't add support for filtering directories at all? Can anyone think of a good api for this?
- [ ] there is an unwrap in the example for `unpack_with_filter()`. Should I make the callback fn return Result<bool> or something?
- [ ] the filter for `append_dir_with_filter()` currently just gives you `path`. I probably also want to know the mtime, to identify files that have changed. I wonder whether there is a point at which the `tar` crate has mtime in its hands but hasn't read the file from disk yet. If so, I could add the filter in there instead.
- [ ] should I be using `filter: impl FnMut(...)` for my function arguments, or does our minimum supported rust version policy forbid this?

If this feature adds too much complexity that you would rather not carry, please close it, and I will take control of the file walking myself in `cargo-quickbuild` instead.

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -3,7 +3,7 @@ use std::cmp;
 use std::convert::TryFrom;
 use std::fs;
 use std::io::prelude::*;
-use std::io::{self, SeekFrom};
+use std::io::{self, Empty, SeekFrom};
 use std::marker;
 use std::path::Path;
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
For unpacking with a filter could that be done by iterating over the entries, filtering, and then using `unpack_in`? It's at least intended that such a method would still be able to leverage the internals of `unpack` ideally.

For `append_dir_with_filter` I'd sort of also prefer to lean on external filters if we can. For example something like `WalkDir` is probably a lot better at filtering what's being walked over rather than this crate.

alsuren:
Okay. Thanks for the feedback. Closing.

Once I have this working in my project, I will come back with a doc-only PR, of the form:

- `Builder::append_dir_all()` -> "If you want more control over what goes into the tarball, use WalkDir and `append_path_with_name()`
- `append_path_with_name()` -> add an example that uses WalkDir, and acts a drop-in replacement for append_dir_all() (I'm assuming that it's okay to add WalkDir as a dev-only dependency, to make the doctests work?)
- `Archive::unpack()` -> "If you want more control over what gets unpacked, iterate over entries in the archive and call Entry::unpack_in()"
- `Entry::unpack_in()` -> remove the `enumerate()` from the example (seems like a copy-paste hangover from the unpack() example).

PR #298, opened 2022-08-12
Add support for PAX Format, Version 1.0

### LINKED ISSUE #286: Unpacking file size is incorrect
And here is the reproduced repo: https://github.com/axetroy/libflate-66

The origin file size is : 985,384
The unzip file size is : 965,416

1. unpack with Tar

size = 965,416

2. unpack with python3

size = 985,384

```py
>>> import tarfile
>>> tar = tarfile.open("prune_darwin_amd64.tar")
>>> tar.getmember("prune").size
985384
```

3. unpack with build-in tar tools in macOS

size = 985,384

ref: https://github.com/sile/libflate/issues/66

### LINKED ISSUE #295: Bug: Extracting tar failed
When extracting [cbindgen-0.24.3-x86_64-apple-darwin.tar.gz](https://github.com/alsuren/cargo-quickinstall/releases/download/cbindgen-0.24.3-x86_64-apple-darwin/cbindgen-0.24.3-x86_64-apple-darwin.tar.gz) using the following equivalent code:

```
let tar = GzDecoder::new(dat);
tar.unpack(path)?;
```

`dat` is some readable that is pipelined from the downloading stage.

Here's the relevant part of the `Cargo.toml`:

```
flate2 = { version = "1.0.24", features = ["zlib-ng"], default-features = false }
tar = "0.4.38"
```

ncihnegn:
To fix #286 #295.


ncihnegn:
I don't understand the macOS/Windows CI failures. Please help.

NobodyXu:
@ncihnegn Thank you very much for this PR!

NobodyXu:
@alexcrichton Can you review this PR please?
It is really important for `cargo-binstall` as users of it encounters PAX format in practice and lack of PAX format means `cargo-binstall` gives errors even if a pre-built binary is available.

src/archive.rs
```diff
@@ -394,26 +396,65 @@ impl<'a> EntriesFields<'a> {
...
     fn parse_sparse_header(&mut self, entry: &mut EntryFields<'a>) -> io::Result<()> {
-        if !entry.header.entry_type().is_gnu_sparse() {
+        if !entry.is_pax_sparse() && !entry.header.entry_type().is_gnu_sparse() {
             return Ok(());
         }
-        let gnu = match entry.header.as_gnu() {
-            Some(gnu) => gnu,
-            None => return Err(other("sparse entry type listed but not GNU header")),
-        };
+        let mut sparse_map = Vec::<SparseEntry>::new();
```
alexcrichton, line 422: One of the main goals I tried to keep for the `tar` crate is to minimize internal allocations. Instead of having a temporary list here could this be refactored to avoid the intermediate allocation?

src/archive.rs
```diff
@@ -256,6 +257,7 @@ impl<'a, R: Read> Iterator for Entries<'a, R> {
     }
 }
 
+#[allow(unused_assignments)] // https://github.com/rust-lang/rust/issues/22630
```
alexcrichton, line 260: I think it would be best to remove this or move the comment to where the warning is printed instead.

src/archive.rs
```diff
@@ -394,26 +396,65 @@ impl<'a> EntriesFields<'a> {
...
+                processed -= 1;
                 continue;
             }
 
             let mut fields = EntryFields::from(entry);
+            fields.pax_extensions = pax_extensions;
+            pax_extensions = None; // Reset pax_extensions after use
+            if is_recognized_header && fields.is_pax_sparse() {
+                gnu_longname = fields.pax_sparse_name();
+            }
```
alexcrichton, line 410: This feels different than the current organization. Instead of pretending that the gnu_longname field was present if a pax-specified field is present could the accessor which looks at `long_pathname` be updated to consult the pax extensions if they're present?

src/archive.rs
```diff
@@ -394,26 +396,65 @@ impl<'a> EntriesFields<'a> {
                 if let Some(pax_extensions_ref) = &pax_extensions {
                     pax_size = pax_extensions_size(pax_extensions_ref);
                 }
+                // Not an entry
+                // Keep pax_extensions for the next ustar header
+                processed -= 1;
```
alexcrichton, line 401: I'm not sure what this is doing? An entry was consumed here so I don't know why this value would be decremented?

src/archive.rs
```diff
@@ -394,26 +396,65 @@ impl<'a> EntriesFields<'a> {
...
+                    .and_then(|s| s.parse::<u64>().ok())
+                    .ok_or_else(|| other("failed to read a decimal line"))
+            };
+
+            let num_entries = read_decimal_line()?;
+            for _ in 0..num_entries {
+                let offset = read_decimal_line()?;
+                let size = read_decimal_line()?;
+                sparse_map.push(SparseEntry { offset, size });
+            }
```
alexcrichton, line 441: I don't think I understand how this could work and pass tests. This is creating a temporary buffer to read from the inner underlying data stream but then the buffer is discarded outside of this scope. That means that more data than necessary could be consumed from the inner data stream and accidentally discarded.

I don't think that this should create a temporary buffer here but instead, if necessary, use a stack-local buffer and then do byte-searching within since presumably the entry here is typically small enough for that.

src/archive.rs
```diff
@@ -394,26 +396,65 @@ impl<'a> EntriesFields<'a> {
...
-        if !entry.header.entry_type().is_gnu_sparse() {
+        if !entry.is_pax_sparse() && !entry.header.entry_type().is_gnu_sparse() {
             return Ok(());
         }
-        let gnu = match entry.header.as_gnu() {
-            Some(gnu) => gnu,
-            None => return Err(other("sparse entry type listed but not GNU header")),
-        };
+        let mut sparse_map = Vec::<SparseEntry>::new();
+        let mut real_size = 0;
```
alexcrichton, line 423: Something about this doesn't feel quite right because `real_size` isn't set in all branches of the `if` statement below whereas prior it was always set to a particular value.

src/archive.rs
ncihnegn: Done.

src/archive.rs
ncihnegn: Done.

src/archive.rs
```diff
@@ -442,13 +483,10 @@ impl<'a> EntriesFields<'a> {
...
             let reader = &self.archive.inner;
             let size = entry.size;
-            let mut add_block = |block: &GnuSparseHeader| -> io::Result<_> {
-                if block.is_empty() {
-                    return Ok(());
-                }
-                let off = block.offset()?;
-                let len = block.length()?;
-                if len != 0 && (size - remaining) % 512 != 0 {
+            let mut add_block = |block: &SparseEntry| -> io::Result<_> {
```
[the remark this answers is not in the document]
ncihnegn: Done.

src/archive.rs
ncihnegn: Done.

src/archive.rs
ncihnegn: It will consume exactly a block of 512 bytes. After reading the necessary data, the remaining filler will be discarded.

src/archive.rs
ncihnegn: I don't think so. Here we need to convert strings to numbers and the number of pairs is not fixed.

akesson:
Did this go stale?

davidanthoff:
Is there any chance that this PR could be revived? Or maybe it is ready to be merged, given that it looks like @ncihnegn responded to all the change requests?

Presumably the first step would be to resolve the merge conflicts, @ncihnegn is that something you could do?

ncihnegn:
Updated.

ararslan:
@alexcrichton, could you please re-review this? The Julia language's version multiplexer/installer uses this crate but we're blocked on porting the installer to all of Julia's supported platforms until this is merged.

alexcrichton:
This is quite an old PR and I've unfortunately lost context on this. I'd also understand if @ncihnegn here wouldn't want to sheperd thing along after it's been 2 years since it's been opened. Despite that though this is somewhat tricky code and I'm hesitant to merge as-is. Merging this would require me to get a better understanding of PAX and how this crate works again (it's been awhile) so that's a fair bit of context to boot back up on. The test coverage also looks relatively light here and additionally there's not a ton of documentation internally about what's going on either.

There's also pieces I don't fully understand like creating intermediate `BufReader` values which feel like they can easily go wrong but I haven't fully investigated them yet. 

Overall I unfortunately do not have the time to myself personally help push this across the finish line, but I can try to outline what would make this easier to land if others are interested in helping to contribute.

cgwalters:
As I understand things, https://github.com/alexcrichton/tar-rs/pull/375 added support for GNU sparse, but this would be the PAX version?

ncihnegn:
> As I understand things, #375 added support for GNU sparse, but this would be the PAX version?
Yes. See 
https://www.gnu.org/software/tar/manual/html_section/Sparse-Formats.html

s-m-e:
@ncihnegn @alexcrichton Hi there - any chance this gets finished / merged? 

I am working my way through adding PAX archive read/write support to a Rust app. If you guys need help, feel free to ping me. I'd love to move/remove this logic out of my app (and have it in a crate instead, ideally this one).

alexcrichton:
I'm strapped pretty thin right now, but @cgwalters or @xzfc would one of y'all be up for helping to push this over the finish line?

src/archive.rs
cgwalters: We could do though:
`let real_size = if entry.is_pax_sparse() { ... } else { ... }`

src/archive.rs
```diff
@@ -430,26 +431,68 @@ impl<'a> EntriesFields<'a> {
...
+                processed -= 1;
                 continue;
             }
 
             let mut fields = EntryFields::from(entry);
-            fields.long_pathname = gnu_longname;
-            fields.long_linkname = gnu_longlink;
             fields.pax_extensions = pax_extensions;
+            // False positive: unused assignment
+            // https://github.com/rust-lang/rust/issues/22630
```
cgwalters, line 443: It looks like this has been fixed so we should be able to drop the assignment.

src/archive.rs
```diff
@@ -282,6 +282,7 @@ impl<'a, R: Read> Iterator for Entries<'a, R> {
     }
 }
 
+#[allow(unused_assignments)]
```
cgwalters, line 285: Hopefully we can drop this now

src/entry.rs
```diff
@@ -300,6 +300,46 @@ impl<'a> EntryFields<'a> {
...
+                    .find(|f| f.key_bytes() == b"GNU.sparse.minor" && f.value_bytes() == b"0")
+                    .is_some();
+        }
+        false
+    }
+
+    pub fn pax_sparse_name(&mut self) -> Option<Vec<u8>> {
+        if let Some(ref pax) = self.pax_extensions {
+            return PaxExtensions::new(pax)
+                .filter_map(|f| f.ok())
```
cgwalters, line 319: I'm not a big fan of "swallowing" errors like this, my preference would be to make this function return `Result<Option<Vec<u8>>>` and propagate this error instead.

src/header.rs
```diff
@@ -124,6 +124,12 @@ pub struct GnuHeader {
     pub pad: [u8; 17],
 }
 
+/// Description of a spare entry.
+pub struct SparseEntry {
+    pub offset: u64,
```
cgwalters, line 129: Let's also document the fields please

src/archive.rs
```diff
@@ -430,26 +431,68 @@ impl<'a> EntriesFields<'a> {
...
-        };
+        let mut sparse_map = Vec::<SparseEntry>::new();
+        let mut real_size = 0;
+        if entry.is_pax_sparse() {
+            real_size = entry.pax_sparse_realsize()?;
+            let mut num_bytes_read = 0;
+            let mut reader = io::BufReader::with_capacity(BLOCK_SIZE as usize, &self.archive.inner);
+            let mut read_decimal_line = || -> io::Result<u64> {
+                let mut str = String::new();
+                num_bytes_read += reader.read_line(&mut str)?;
```
cgwalters, line 468: Hmm can't we just call `reader.lines().next()`?

cgwalters (changes requested):
I just did a mostly superficial pass so far, haven't looked at correctness of sparse handling (I did glance at [the gnu tar docs](https://www.gnu.org/software/tar/manual/html_section/Sparse-Formats.html) a while ago and just recall it's a mess)

src/archive.rs
ncihnegn: Dropping comments only.

src/archive.rs
ncihnegn: No, rustc 1.87.0-nightly still complains.

src/archive.rs
ncihnegn: We don't have extra assignments here.

src/header.rs
ncihnegn: Offset and size names are self explanatory.

src/archive.rs
ncihnegn: What's the benefit? `Lines.next()` is `Option<Result<String>>`, which needs extra handling.

src/entry.rs
ncihnegn: I don't understand. To propagate the error using `Result` we will be dropping the `Vec`.

s-m-e:
Ping.

s-m-e:
Yet another ping - what's still left to do to get this one over the finish line? I've wrapped good old GNU tar in my app, and its about as horrible as one might expect. I'd rather help to get this PR done.

yjqg6666:
Any update for this PR?

PR #300, opened 2022-08-22
Fix Windows CI

ncihnegn:


alexcrichton:
Personally I don't want to get in to the business of interpreting paths too much since it's already tricky enough. Stripping a leading `//` already feels quite dubious. Is there perhaps a way to fix the test or otherwise test for a different result on Windows? Or use different `Path` APIs on Windows?

ncihnegn:
Are you open to use a crate for ```UnixPath```?

alexcrichton:
I'm not personally familiar with any such crates, is there no way to do this using the standard library?

ncihnegn:
I think using ```UnixPath``` is the cleanest way.
See https://github.com/rust-lang/rust/issues/100833.

ncihnegn:
Simplified a bit without using a crate; please review.

PR #301, opened 2022-09-22
Emscripten support

xzfc:
The target `wasm32-unknown-emscripten` is considered both `wasm32` and `unix`, so the current `master` branch does not compile with multiple definition errors. This PR fixes this. Some of the tests checking file mtimes and symlinks still fail on Emscripten, this could be addressed later.

Relies on https://github.com/alexcrichton/filetime/pull/88.


alexcrichton (commented):
I'm a bit worried about the growth of the usage of `#[cfg]` here, while windows/unix is pretty simple adding windows/unix/wasm/not-emscripten is getting a bit much. Would you be willing to integrate the `cfg-if` crate to clean up these definitions?

alexcrichton (approved):
Thanks!

atilag:
I'm trying to port a project to `wasm32-unknown-emscripten` that has this crate as a transient dependency. Willing to take over this and finish the emscripten port just in case @xzfc or @alexcrichton don't have the time to do it :)

atilag:
I've been taking a look at the PR, and it's becoming challenging to read between different `if #[cfg()] / else`.  For short functions it's not a problem, though. [Cognitive Complexity](https://docs.codeclimate.com/docs/cognitive-complexity) increases significantly with that extra indentation level :)

ndmitchell:
I'm having this problem too. FWIW, the way I hacked around this was by replacing `#[cfg(unix)]` with `#[cfg(all(unix, not(target_arch = "wasm32")))]` in a handful of places. That seemed fairly self-contained and not too horrible. I'd be happy to spin up a PR, or revive this one if that was what was deemed appropriate. Let me know how I can help @alexcrichton  😄

alexcrichton:
This is a pretty old PR at this point clocking in at ~2.5 years, so yeah if you're interested I think it'd be best to probably start with a new PR and go from there?

ndmitchell:
Updated as #397

Issue #313, opened 2023-03-28
error: "numeric field was not a number: t Regard when getting cksum for ...."


npajkovsky:
Hello,

I have a tarball that I want to read
```
-> % file backup.tar
backup.tar: POSIX tar archive
```

But I've got an error
```
error: "numeric field was not a number: t Regard when getting cksum for a, <=\nm>\n> wrote:\n>\n> Dear \n>\n>\n>\n> Correct. We will have more detai"
```

It looks like that it starts reading a header from a wrong position.

My code looks like this
```
use std::fs::File;

use eyre::Result;
use tar::Archive;

fn main() -> Result<()> {
    let file: File = File::open(path: "backup.tar")?;
    let mut archive: Archive<File> = Archive::new(obj: file);

    archive.entries().unwrap().for_each(|x: Result<Entry<File>, Error>| {
        let x: Entry<File> = x.unwrap();
        println!("{}", x.path().unwrap().display())
    });

    Ok(())
}
```
And with a couple of dbg!() I've got this.
```
[/home/nikola/src/tar-rs/src/archive.rs:268] self.next = 1081462272
[/home/nikola/src/tar-rs/src/archive.rs:271] self.next - self.archive.inner.pos.get() = 7680
[/home/nikola/src/tar-rs/src/archive.rs:294] &header = UstarHeader {
    entry_size: 138,
    size: 138,
    path: "././@LongLink",
    link_name: None,
    username: Some(
        "",
    ),
    groupname: Some(
        "",
    ),
    cksum: 2567,
    cksum_valid: true,
}
[/home/nikola/src/tar-rs/src/archive.rs:268] self.next = 1081463296
[/home/nikola/src/tar-rs/src/archive.rs:271] self.next - self.archive.inner.pos.get() = 374
[/home/nikola/src/tar-rs/src/archive.rs:294] &header = UstarHeader {
    entry_size: 64700,
    size: 64700,
    path: "backup-3.22.2023_17-45-34_acaegypt/homedir/mail/....t/management/cur/1673424346.M756912P292403.demosthenes.datapack.net,S=64700,W=65570:2,",
    link_name: Some(
        "backup-3.22.2023_17-45-34_acaegypt/homedir/mail/...t/acc/cur/1673424346.M756912P292403.demos",
    ),
    mode: 0o640,
    uid: 1058,
    gid: 1061,
    mtime: 1673424346,
    username: Some(
        "user",
    ),
    groupname: Some(
        "group",
    ),
[...]

PR #314, opened 2023-03-29
header: set entry_size() to 0 for hardlinks

npajkovsky:
Fix the error: "numeric field was not a number: t 'Regard' when getting cksum for 'a text ...'. The error is caused by wrongly setting the `self.next` position due to accounting harlinks size.

According to man 5 tar, the size of the hard-links should be set to 0.

size    Size of file, as octal number in ASCII.  For regular files	only,
        this indicates the	amount of data that follows the	header.	 In
        particular, this field was	ignored	by early tar implementations
        when extracting hardlinks.	 Modern	writers	should always store a
        zero length for hardlink entries.

But since the writer wasn't *modern*, the entry_size is 64700 which causes miscalculation of the `self.next`.

[tar-rs/src/archive.rs:372] &entry.header() = UstarHeader {
    entry_size: 64700,
    size: 64700,
    path: "some/path",
    link_name: Some(
        "some/link",
    ),
    mode: 0o640,
    uid: 1058,
    gid: 1061,
    mtime: 1673424346,
    username: Some(
        "example",
    ),
    groupname: Some(
        "example",
    ),
    device_major: Some(
        9,
    ),
    device_minor: Some(
        2,
    ),
    cksum: 24700,
    cksum_valid: true,
}
[tar-rs/src/archive.rs:373] entry.header().entry_type() = Link

Closes: #313

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -298,6 +298,9 @@ impl Header {
     ///
     /// May return an error if the field is corrupted.
     pub fn entry_size(&self) -> io::Result<u64> {
+        if self.entry_type().is_hard_link() {
+            return Ok(0);
+        }
         num_field_wrapper_from(&self.as_old().size).map_err(|err| {
             io::Error::new(
                 err.kind(),
--- a/tests/header/mod.rs
+++ b/tests/header/mod.rs
@@ -1,11 +1,13 @@
-use std::fs::{self, File};
-use std::io::{self, Write};
-use std::path::Path;
-use std::{iter, mem, thread, time};
+use std::{
+    fs::{self, File},
+    io::{self, Write},
+    iter, mem,
+    path::Path,
+    thread, time,
+};
 
+use tar::{EntryType, GnuHeader, Header, HeaderMode};
 use tempfile::Builder;
-
-use tar::{GnuHeader, Header, HeaderMode};
 
 #[test]
 fn default_gnu() {
@@ -243,4 +245,15 @@ fn byte_slice_conversion() {
     let b: &[u8] = h.as_bytes();
     let b_conv: &[u8] = Header::from_byte_slice(h.as_bytes()).as_bytes();
     assert_eq!(b, b_conv);
+}
+
+#[test]
+fn hardlink_entry_size() {
+    let mut h = Header::new_ustar();
+    let p = Path::new("a").join(&vec!["a"; 100].join(""));
+    h.set_entry_type(EntryType::Link);
+    t!(h.set_path(&p));
+    t!(h.set_link_name("foo"));
+    h.set_size(200);
+    assert_eq!(t!(h.entry_size()), 0);
 }


alexcrichton:
If the source of this is the man page then I would interpret that as the archive in question being an invalid archive rather than one that should be silently handled to work instead. Does the `tar` CLI, though, handle this case? Would you be able to find in its source code where it handles this case?

npajkovsky:
Both bsdtar and tar handles this tarball correctly.

npajkovsky:
libarchive does this

`libarchive/libarchive/archive_read_support_format_tar.c`
```
        /*
          * The following may seem odd, but: Technically, tar
          * does not store the file type for a "hard link"
          * entry, only the fact that it is a hard link.  So, I
          * leave the type zero normally.  But, pax interchange
          * format allows hard links to have data, which
          * implies that the underlying entry is a regular
          * file.
          */
         if (archive_entry_size(entry) > 0)
             archive_entry_set_filetype(entry, AE_IFREG);

         /*
          * A tricky point: Traditionally, tar readers have
          * ignored the size field when reading hardlink
          * entries, and some writers put non-zero sizes even
          * though the body is empty.  POSIX blessed this
          * convention in the 1988 standard, but broke with
          * this tradition in 2001 by permitting hardlink
          * entries to store valid bodies in pax interchange
          * format, but not in ustar format.  Since there is no
          * hard and fast way to distinguish pax interchange
          * from earlier archives (the 'x' and 'g' entries are
          * optional, after all), we need a heuristic.
          */
         if (archive_entry_size(entry) == 0) {
             /* If the size is already zero, we're done. */
         }  else if (a->archive.archive_format
             == ARCHIVE_FORMAT_TAR_PAX_INTERCHANGE) {
             /* Definitely pax extended; must obey hardlink size. */
         } else if (a->archive.archive_format == ARCHIVE_FORMAT_TAR
             || a->archive.archive_format == ARCHIVE_FORMAT_TAR_GNUTAR)
         {
             /* Old-style or GNU tar: we must ignore the size. */
             archive_entry_set_size(entry, 0);
             tar->entry_bytes_remaining = 0;
         } else if (archive_read_format_tar_bid(a, 50) > 50) {
             /*
              * We don't know if it's pax: If the bid
              * function sees a valid ustar header
              * immediately following, then let's ignore
              * the hardlink size.
              */
             archive_entry_set_size(entry, 0);
             tar->entry_bytes_remaining = 0;
         }
         /*
          * TODO: There are still two cases I'd like to handle:
          *   = a ustar non-pax archive with a hardlink entry at
          *     end-of-archive.  (Look for block of nulls following?)
          *   = a pax archive that has not seen any pax headers
          *     and has an entry which is a hardlink entry storing
          *     a body containing an uncompressed tar archive.
          * The first is worth addressing; I don't see any reliable
          * way to deal with the second possibility.
          */
         break;
```

alexcrichton:
Thanks! I'm not sure if those line up though? Your second link unconditionally sets the size to zero, and the first code you copy/pasted only sets the size to zero for non-ustar archives. The PR, however, unconditionally sets it to zero. Is the copy/pasted code outdated or otherwise no longer in use?

npajkovsky:
The copy/pasted was copied directly from the libarchive git repo HEAD. I think the interesting part is 
```
} else if (archive_read_format_tar_bid(a, 50) > 50) {
             /*
              * We don't know if it's pax: If the bid
              * function sees a valid ustar header
              * immediately following, then let's ignore
              * the hardlink size.
              */
             archive_entry_set_size(entry, 0);
             tar->entry_bytes_remaining = 0;
         }
```

PR #317, opened 2023-04-13
archive: work from leaves up when unpacking directories

jeckersb:


drahnr:
Can be closed, https://github.com/alexcrichton/tar-rs/pull/372 implements the same thing

PR #328, opened 2023-06-20
tests: Don't assume /dev/loop0 exists

cgwalters:
It doesn't by default on e.g. current Fedora and RHEL systems. I suspect most people so far have been running this on Ubuntu which uses snaps which use loopback devices, but we can't rely on that.

For test coverage here, ultimately there's not really a good block device to rely on without going through a lot of special casing.

Let's just test character devices.


vrmiguel:
+1 since this also does not work on Arch Linux

alexcrichton:
Thanks!

PR #329, opened 2023-06-21
entry: Use `strip_prefix`

cgwalters:
Minor drive by code cleanup; `strip_prefix` is just more readable and more bug-resistant than manually doing it with `starts_with` combined with slicing.

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -841,11 +841,7 @@ impl<'a> EntryFields<'a> {
                 .filter_map(|e| {
                     let key = e.key_bytes();
                     let prefix = b"SCHILY.xattr.";
-                    if key.starts_with(prefix) {
-                        Some((&key[prefix.len()..], e))
-                    } else {
-                        None
-                    }
+                    key.strip_prefix(prefix).map(|rest| (rest, e))
                 })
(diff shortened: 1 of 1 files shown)


cgwalters:
@alexcrichton This crate is currently very important to some software I maintain; I'd probably be able to act as a 20%-time co-maintainer.  Maybe what we could do is find a few other interested parties, and switch things so that merging a PR requires 2 approving reviews?  Looking at the outstanding PRs there are some other people in the list that I'd trust to be co-maintainers too, e.g. @sunshowers.

alexcrichton:
Thank you for the offer! I would be happy to have help maintaining this crate as I think I clearly can't offer the support it needs.

If y'all would like would you be ok if I sort of "phased y'all in" over time? For example would you be ok reviewing PRs and then pinging me when it's ready to merge? After a few PRs I can give y'all write access to merge PRs yourselves and then you'd only need to ping me for releases. After a release or two perhaps I could invite y'all to crates.io to publish without me as well.

alexcrichton:
All looks great, thank you again for your help here, it's very much appreciated! Feel free to ping me for merging those PRs when they're ready and I'll do a final once-over, your comments look good to me.

It's also been awhile since a release has been made, if you're up for it mind opening a PR with a version bump? I can publish to crates.io after when it's ready.

PR #330, opened 2023-06-29
feat: setting permission mask in `Entry` and `Archive`

weihanglo:
This is convenience when you want to disable some permission bits
during unpacking. Currently only support Unix.

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -22,6 +22,7 @@ pub struct Archive<R: ?Sized + Read> {
 
 pub struct ArchiveInner<R: ?Sized> {
     pos: Cell<u64>,
+    mask: u32,
     unpack_xattrs: bool,
     preserve_permissions: bool,
     preserve_ownerships: bool,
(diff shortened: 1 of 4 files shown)


weihanglo:
Another solution is that we can have a `set_mask()` function to set mask. It might also be useful when a user just wants to remove some permissions.

joshtriplett:
@weihanglo I would suggest splitting the UNC path fix into a separate PR?

And I do think mask would be more convenient than mode, since then tar can still handle "executable or not" itself.

weihanglo:
Thanks @joshtriplett. #331 has been created for UNC path fix.

And this PR is also switched to `set_mask` from `set_mode`. I am not sure about how that works on Windows though.

alexcrichton:
Thanks! Right now the mask looks like it's inverted when applying against a mode, so could that be documented? Either that or could the mask not be inverted and the default mask is `u32::MAX`?

weihanglo:
Thanks for the review! I went ahead with a logical notation. If you like not inverted one please let me know :)

alexcrichton (approved):

alexcrichton:
Sorry for the delay, but sure yeah published now.

PR #337, opened 2023-08-28
Add a force_mtime() function to allow overriding all mtime fields

grahamc:
Hi Alex,

Thanks for the excellent tar crate. While developing a project, we needed a way to override the mtime. This was the most direct route to get it done, even though I'm rather sure it isn't something you'd want to merge upstream.

So, here is some code we wrote. We're sending the PR for you to look at and consider as a use case. We'd be interested in hearing back from you if you're interested in a better implementation, and if so what you think that might look like.

Thank you!

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -16,6 +16,7 @@ pub struct Builder<W: Write> {
     follow: bool,
     finished: bool,
     obj: Option<W>,
+    force_mtime: Option<u64>,
 }
 
 impl<W: Write> Builder<W> {
@@ -28,6 +29,7 @@ impl<W: Write> Builder<W> {
             follow: true,
             finished: false,
             obj: Some(obj),
+            force_mtime: None,
         }
     }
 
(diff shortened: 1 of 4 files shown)


adamqqqplay (commented):
@grahamc Hi grahamc, it looks like the CI test is failing, maybe you could fix it first, thanks!

grahamc:
We certainly could. Does that mean you're interested in accepting this change as designed?

cgwalters:
Hi @grahamc :wave: 

This relates a lot to e.g. https://github.com/alexcrichton/tar-rs/pull/346 and https://github.com/alexcrichton/tar-rs/pull/343 a bit.

In your use case, do you *also* want to canonicalize uid/gid?  I'd guess so...which would argue that we should fix `HeaderMode::Deterministic`.

Also I think we should call out that it's really not terribly hard to do "manually" walk a directory recursively and add entries to it:

See e.g. https://github.com/coreos/cargo-vendor-filterer/blob/28c7760484240184c95f76b5a3f11d92a6779ff1/src/main.rs#L475-L510

We could put that in `examples/build-walkdir.rs` or something that people could at least copy-paste as a reference?

PR #343, opened 2023-11-02
feat: allow to skip unsupported files

yet-another-aligator:
This PR provide a new `Builder::skip_unsupported_file_types(&mut self, skip: bool)` method, for allowing to silently skip unsupported files, instead of returning with an error.

For the moment, this feature applies to the case of UNIX named sockets only.


cgwalters (commented):
I think we really just need this as a variant of `append_dir_all()` right?  I'd probably have it take a closure which filters the entries.

PR #346, opened 2023-11-13
Unify `mtime` constant used on Unix and Windows

### LINKED ISSUE #341: HeaderMode::Deterministic differs in behaviour on Unix and Windows platforms
On Windows the `mtime` is set to the dummy value of https://github.com/alexcrichton/tar-rs/blob/master/src/header.rs#L828

```rust
                self.set_mtime(123456789); // see above in unix
```
but on Unix it is set to https://github.com/alexcrichton/tar-rs/blob/master/src/header.rs#L760

```rust
                // We could in theory set the mtime to zero here, but not all
                // tools seem to behave well when ingesting files with a 0
                // timestamp. For example rust-lang/cargo#9512 shows that lldb
                // doesn't ingest files with a zero timestamp correctly.
                //
                // We just need things to be deterministic here so just pick
                // something that isn't zero. This time, chosen after careful
                // deliberation, corresponds to Jul 23, 2006 -- the date of the
                // first commit for what would become Rust.
                self.set_mtime(1153704088);
```

I don't see why these should be different, and it causes an annoyance in one of our use-cases.

Is there a reason why these two dummy values should be different?

mkaput:
fix #341

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -16,6 +16,14 @@ use std::str;
 use crate::other;
 use crate::EntryType;
 
+/// A deterministic, arbitrary, non-zero timestamp that use used as `mtime`
+/// of headers when [`HeaderMode::Deterministic`] is used.
+///
+/// This value, chosen after careful deliberation, corresponds to _Jul 23, 2006_,
+/// which is the date of the first commit for what would become Rust.
+#[cfg(any(unix, windows))]
+const DETERMINISTIC_TIMESTAMP: u64 = 1153704088;
+
 /// Representation of the header of an entry in an archive
 #[repr(C)]
 #[allow(missing_docs)]
@@ -748,16 +756,11 @@ impl Header {
                 self.set_mode(meta.mode() as u32);
             }
             HeaderMode::Deterministic => {
-                // We could in theory set the mtime to zero here, but not all
-                // tools seem to behave well when ingesting files with a 0
-                // timestamp. For example rust-lang/cargo#9512 shows that lldb
-                // doesn't ingest files with a zero timestamp correctly.
-                //
-                // We just need things to be deterministic here so just pick
-                // something that isn't zero. This time, chosen after careful
-                // deliberation, corresponds to Jul 23, 2006 -- the date of the
-                // first commit for what would become Rust.
-                self.set_mtime(1153704088);
+                // We could in theory set the mtime to zero here, but not all tools seem to behave
+                // well when ingesting files with a 0 timestamp.
+                // For example, rust-lang/cargo#9512 shows that lldb doesn't ingest files with a
+                // zero timestamp correctly.
+                self.set_mtime(DETERMINISTIC_TIMESTAMP);
 
                 self.set_uid(0);
                 self.set_gid(0);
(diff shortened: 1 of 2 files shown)


cgwalters (changes requested):
There's clearly no reason for them to be different.  However, I do have a lingering concern here that some projects may be e.g. checksumming the result of the tar stream produced by `HeaderMode::Deterministic` and so this could break them.

I'm personally tentatively OK with changing this, but can you hoist this to a `const ARBITRARY_DETERMINISTIC_TIMESTAMP` that is explicitly shared?

We may also want to consider e.g.

```
#[cfg(windows)]
pub const PREVIOUS_DETERMINISTIC_TIMESTAMP: <old value>
#[cfg(unix)
pub const PREVIOUS_DETERMINISTIC_TIMESTAMP: <current value>
```
or so so that anyone who is bit by this can explicitly still use `HeaderMode::Deterministic` but then also override the time back to the previous value?

cgwalters (approved):

mkaput:
> I'm personally tentatively OK with changing this, but can you hoist this to a `const ARBITRARY_DETERMINISTIC_TIMESTAMP` that is explicitly shared?

Done



> We may also want to consider e.g.
> 
> ```
> #[cfg(windows)]
> pub const PREVIOUS_DETERMINISTIC_TIMESTAMP: <old value>
> #[cfg(unix)
> pub const PREVIOUS_DETERMINISTIC_TIMESTAMP: <current value>
> ```
> 
> or so so that anyone who is bit by this can explicitly still use `HeaderMode::Deterministic` but then also override the time back to the previous value?

I feel uneasy with this because:
1. This makes something that used to be a bug, soon to be fixed, a public API (?) Looks contradictory.
2. If somebody heavily relied on these values, then this means they made them part of their public interface (i.e. they leaked `tar-rs` implementation detail as their public thing). So this looks like a problem of the user that `tar-rs` should not worry about?
3. This change is worthy of a breaking release anyway, I believe?
4. This crate also supports other targets (wasm32), for which such constant does not exist.

mkaput:
For context, that's how I reacted to this problem in the project I stumbled upon this:
https://github.com/software-mansion/scarb/blob/fc81b059a81d164c88eaed021a7b84c63fee5d00/scarb/src/ops/package.rs#L456-L466

cgwalters:
Right.  Actually if we were breaking semver I'd argue `HeaderMode::Deterministic` should become `HeaderMode::Deterministic { mtime: u64 }` - and callers should pass their own timestamp.  This is e.g. what I do in https://github.com/coreos/cargo-vendor-filterer/blob/28c7760484240184c95f76b5a3f11d92a6779ff1/src/main.rs#L492 which ultimately takes the timestamp from the git commit, as is recommended by https://reproducible-builds.org/docs/source-date-epoch/

mkaput:
> Right. Actually if we were breaking semver I'd argue `HeaderMode::Deterministic` should become `HeaderMode::Deterministic { mtime: u64 }` - and callers should pass their own timestamp. This is e.g. what I do in https://github.com/coreos/cargo-vendor-filterer/blob/28c7760484240184c95f76b5a3f11d92a6779ff1/src/main.rs#L492 which ultimately takes the timestamp from the git commit, as is recommended by https://reproducible-builds.org/docs/source-date-epoch/

That sounds like the smartest move because it would force people to reason about that timestamp.

alexcrichton:
Thanks for the PR (and sticking it out through the delay @mkaput) and thanks for the review @cgwalters!

PR #348, opened 2023-12-12
update xattr to 1.1.3

sunshowers:
xattr 1.0.0 through 1.1.2 have a bug
(https://github.com/Stebalien/xattr/pull/48). Bump the requirement so that downstream dependents get the latest version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ filetime = "0.2.8"
 tempfile = "3"
 
 [target."cfg(unix)".dependencies]
-xattr = { version = "1.0", optional = true }
+xattr = { version = "1.1.3", optional = true }
 libc = "0.2"
 
 [features]


adamqqqplay:
@sunshowers Thanks! Currently xattr has released v1.3.1, can we bump to this version?

sunshowers:
Sure, note that I'm not a maintainer so you may wish to write your own PR.

alexcrichton:
Thanks!

I'll also note that technically this isn't necessary since these versions are all "1.x.x" compatible meaning that downstream consumers can always update. This change works by preventing `tar` from being used with previous versions (with the next release of `tar`)

alexcrichton:
Also sorry for the delay on merging :(

sunshowers:
Thanks! The problem is not that downstream dependents can't update, it's that they won't update transitive deps unless prodded.

PR #363, opened 2024-06-06
Remove unused `use` in doctest

schneems:
The `use std::fs` in the for `append_dir_all` tests is not used. We can simplify this example by removing it.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -365,7 +365,6 @@ impl<W: Write> Builder<W> {
     /// # Examples
     ///
     /// ```
-    /// use std::fs;
     /// use tar::Builder;
     ///
     /// let mut ar = Builder::new(Vec::new());


src/builder.rs
```diff
@@ -365,7 +365,6 @@ impl<W: Write> Builder<W> {
     /// # Examples
     ///
     /// ```
-    /// use std::fs;
```
schneems, line 368: I also saw that Clippy isn't running in CI, and there are quite a few warnings: https://gist.github.com/schneems/fd26378009f9ac28ef282226de6ba5e7. Before going overboard, I wanted to send this one small change. In searching history to see if that's intentional, I saw https://github.com/alexcrichton/tar-rs/pull/79#discussion_r71975483 where you prefer a different style than Clippy suggests (which makes me feel that you would be opposed to adding a clippy check to CI).

PR #364, opened 2024-06-06
Update append_dir_all doc test

schneems:
## Call `finish`

The text of `append_dir_all` suggests that `finish` must always be called. This commit updates the example to call `finish`.

## Link to `into_inner`

Further, the docs for `finish` suggest that `into_inner` is an alternate (and possibly preferred), so I added a link to that function.

> /// In most situations the `into_inner` method should be preferred.
>     pub fn finish(&mut self) -> io::Result<()> {

## Label which directory is which in the comment

Lastly, I needed clarification about which directory was renamed in the example. Without prior knowledge, it could be that a directory named "bardir" is being renamed to the current dir or that the current dir is being inserted as a "bardir." The input args are `path` and `src_path`, which provides a little clue. 

To make the example more explicit, I added that information to the comment.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -360,7 +360,8 @@ impl<W: Write> Builder<W> {
     /// operation then this may corrupt the archive.
     ///
     /// Also note that after all files have been written to an archive the
-    /// `finish` function needs to be called to finish writing the archive.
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):

alexcrichton:
Thanks for this!

PR #365, opened 2024-06-06
Add append_dir_all doc test without renaming

schneems:
This adds a doc test that shows how to add the contents of a directory to an archive without renaming them.

## Context

It might seem silly, but I was unsure of the correct way to write all the contents of a directory to an archive without renaming. I initially looked for a different associated function that only took one argument but couldn't find one. When that failed, I guessed and checked that `"."` would work, and it did. On searching, others used `""`. 

It makes sense in hind-site, but I would have loved to have a signal that this was the correct way to perform that operation.

In addition to demonstrating `append_dir_all("", ...` this example also shows usage of `into_inner` to retrieve a reference to the original struct. Which is the recommended alternative to `finish`.

I certainly understand that you can't have an example for every permutation and combination, but I believe that archiving the contents of a directory without renaming is common-enough to warrant an explicit example.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -374,6 +374,35 @@ impl<W: Write> Builder<W> {
     /// // with a different name.
     /// ar.append_dir_all("bardir", ".").unwrap();
     /// ```
+    ///
+    /// Use `append_dir_all` with an empty string as the first path argument to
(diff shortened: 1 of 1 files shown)


alexcrichton (approved):
Thanks!

PR #368, opened 2024-06-30
builder: Note that `append_dir_all` doesn't hardlink

cgwalters:
I got caught out by this not generating hardlinks. At this point it's probably part of our "ABI", so just add a note in the docs.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -362,6 +362,11 @@ impl<W: Write> Builder<W> {
     /// Also note that after all files have been written to an archive the
     /// `finish` function needs to be called to finish writing the archive.
     ///
+    /// # Hardlinks
+    ///
+    /// This function will never create hardlinks; to do that, you will
+    /// need to implement a custom recursive walk.
+    ///
     /// # Examples
     ///
     /// ```


src/builder.rs
```diff
@@ -362,6 +362,11 @@ impl<W: Write> Builder<W> {
     /// Also note that after all files have been written to an archive the
     /// `finish` function needs to be called to finish writing the archive.
     ///
+    /// # Hardlinks
+    ///
+    /// This function will never create hardlinks; to do that, you will
+    /// need to implement a custom recursive walk.
+    ///
```
schneems, line 369: Not the maintainer (thanks for the review of my PRs!). I'm also a newb when it comes to a lot of the details of tar implementations and details. Leaving a comment here so you can dismiss/hide it and it won't take up space on the main thread.

1. What is the behavior when it encounters hardlinks? Will it copy them like https://github.com/alexcrichton/tar-rs/pull/366? Would it make sense to spell out what "never create[ing] hardlinks" means in practice? I would appreciate that info when viewing the docs.
2. Do you know if that behavior is intentional or accidental? I.e. Could this perhaps be a bug to be fixed rather than a behavior to be documented? Genuine question.
3. If it's intentional, could adding a method to the builder like `preserve_hardlinks(bool)` be a way forward?

My goal here is to better understand the library and its existing behavior and to explore and understand the possibility of ways we could improve this library or possibly the PR.

PR #371, opened 2024-08-07
[Symlink] Add symlink ownership management.

stswidwinski:
## Context 

The `tar-rs` crate ignores ownership of symlinks, even if `.set_preserve_ownerships(true);` is invoked. While permissions of symlinks are mostly meaningless on most systems, ownership does dictate the access to symlinks. For instance:

```
$ mkdir foo 
$ sudo ln -s foo link
$  rm link
rm: link: Permission denied
```

## This PR

In this PR I add the right invocation for symlink ownership change and expand the tests to include directories and symlinks.

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -576,6 +576,12 @@ impl<'a> EntryFields<'a> {
                             ),
                         )
                     })?;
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):
Thanks for this!

PR #372, opened 2024-08-19
Ensure directory permissions are correctly applied.

stswidwinski:
The current implementation will fail on directory structures which have parent directories less permissive than child directories. For instance, any structure of the sort:

```
- root_dir (0o555)
  - child_dir (0o755)
```

Will fail to be created since `root_dir` permissions won't allow `child_dir` to be created. The simple fix here is to reorder the directories and treat the filesystem as a tree processed in topological order. This PR does just that.

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -238,6 +238,15 @@ impl Archive<dyn Read + '_> {
                 file.unpack_in(dst)?;
             }
         }
+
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):

alexcrichton:
Thanks!

PR #374, opened 2024-08-22
fix: Don't forget to zero-terminate in octal_into

xzfc:
Usually, numeric fields like `size` are set only once, implicitly assuming that the previous data is zero-initialized.  However, it's not always the case, particularly when overriding large values written in the extended form.


alexcrichton (approved):
Thanks!

PR #376, opened 2024-09-05
feat: Add Builder::append_writer

### LINKED ISSUE #304: Alternative append API
At the moment in the API, I have to provide a `Read` and then the `append` function(s) will `io::copy` over the bytes in one go. This is fine for many use cases, but not always desirable.

As an alternative that may work better in some cases, I propose having an API that instead of taking a `Read` returns a `Write`. One can then write into there in order to write into the archive. Lifetimes are use to enforce that no other functions can be called on the archive in the meantime, and the `Drop` implementation takes care of finalizing the file (i.e. add the padding).

xzfc:
This PR adds a new method `Builder::append_writer` which returns `EntryWriter` which implements `Write`.

Closes #304 (more details there).


src/builder.rs
```diff
@@ -440,6 +481,92 @@ impl<W: Write> Builder<W> {
...
+
+    /// Finish writing the current entry in the archive.
+    pub fn finish(self) -> io::Result<()> {
+        let mut this = std::mem::ManuallyDrop::new(self);
+        this.do_finish()
+    }
+
+    fn do_finish(&mut self) -> io::Result<()> {
+        // Pad with zeros if necessary.
+        let buf = [0; 512];
```
cgwalters, line 533: I think type inference handles this but I'd explicitly spell out `0u8` anyways

cgwalters (commented):
Looks sane to me, one superficial nit

src/builder.rs
xzfc: Updated.

alexcrichton (approved):
Also looks great to me, thank you for this!

src/builder.rs
```diff
@@ -440,6 +481,92 @@ impl<W: Write> Builder<W> {
...
+        Ok(EntryWriter {
+            obj,
+            header,
+            written: 0,
+        })
+    }
+
+    /// Finish writing the current entry in the archive.
+    pub fn finish(self) -> io::Result<()> {
+        let mut this = std::mem::ManuallyDrop::new(self);
```
cgwalters, line 527: I know I'm commenting on a closed PR...I meant to check this at the time but recent activity here reminded me to follow up.

Isn't this just going to always leak memory? When I first saw this I thought this pattern I hadn't seen before was actually an elegant way to handle "fallible drops" (in other places I've just put an `Option<T>` that `do_finish()` would `.take()` leaving the `Drop` to be a no-op)

But...so I just typed out this code

```
diff --git a/tests/all.rs b/tests/all.rs
index eff0e4d..87f94ea 100644
--- a/tests/all.rs
+++ b/tests/all.rs
@@ -4,11 +4,13 @@ extern crate tempfile;
 #[cfg(all(unix, feature = "xattr"))]
 extern crate xattr;
 
+use core::time;
 use std::fs::{self, File};
 use std::io::prelude::*;
 use std::io::{self, Cursor};
 use std::iter::repeat;
 use std::path::{Path, PathBuf};
+use std::thread;
 
 use filetime::FileTime;
 use tar::{Archive, Builder, Entries, EntryType, Header, HeaderMode};
@@ -1681,3 +1683,32 @@ fn pax_and_gnu_uid_gid() {
         }
     }
 }
+
+struct Foo<'a> {
+    s: &'a str,
+    v: Vec<u8>,
+}
+
+impl<'a> Foo<'a> {
+    fn finish(self) -> io::Result<()> {
+        let mut this = std::mem::ManuallyDrop::new(self);
+        this.do_finish()
+    }
+
+    #[inline(never)]
+    fn do_finish(&mut self) -> io::Result<()> {
+        std::hint::black_box(Ok(()))
+    }
+}
+
+#[test]
+fn test_manually_drop() {
+    for _ in 0..1_000_000 {
+        let v = Foo {
+            s: "test",
+            v: vec![0u8; 1_000_000],
+        };
+        v.finish().unwrap();
+        std::thread::sleep(time::Duration::from_millis(100));
+    }
+}
```

(I use one-off test functions as basically a local play.rust-lang.org) and watching the cargo test run indeed I see the process memory continually growing.

Can you explain - am I missing something here? If not...it seems like you must have picked up or used this `std::mem::ManuallyDrop` pattern elsewhere and those places need to be fixed.

cgwalters: Ah never mind duh, of course it's going to leak if the internal types have drop glue like `Box` but I guess it doesn't if they don't?

Hmm...it seems like this one could at least use a comment explaining why it's not leaking. I guess what'd make this pattern more robust is something like a ```
#[deny(impl Drop)]
struct EntryWriter {
```

I don't know if there's crate or a builtin way to do that today. Maybe the best followup is a note in the `ManuallyDrop` docs about this?

alexcrichton: Ah good point! And yeah AFAIK there's no great way to lint this right now, so a comment sounds like the best course of action to me.

cgwalters: PR in https://github.com/alexcrichton/tar-rs/pull/380

xzfc: If there was a clean way to express "deconstruct this structure, but do not call Drop", this part of code would be clearer. But it's not possible in the current Rust without hacks.

PR #377, opened 2024-09-11
builder: Add a set_rdev() convenience

cgwalters:
Commonly one may be building tar archives from the filesystem in a custom way, or operating on a translation of a non-tar format which uses rdev, not split major/minor.

Today one needs to replicate the nontrivial bit manipulations to split the device; but since we have that code anyways in our own "build from filesystem" code, let's just add it as a public API.


cgwalters:
Hum right, `dev_t` here is platform dependent. Fun. OK I switched to `libc::dev_t` in the public API.

alexcrichton:
Took me awhile to get here, but seems reasonable to me! I think there's still some compilation issues though?

cgwalters:
Yeah this needs some more work; not at all urgent. I'll unmark it as draft when ready.

PR #380, opened 2024-09-26
EntryWriter: Add comments for why ManuallyDrop is OK

cgwalters:
This wasn't immediately obvious to me at least.


src/builder.rs
```diff
@@ -502,6 +502,8 @@ impl<T: Write + Seek> SeekWrite for T {
 /// After writing all data to the entry, it must be finalized either by
 /// explicitly calling [`EntryWriter::finish`] or by letting it drop.
 pub struct EntryWriter<'a> {
+    // NOTE: Do not add any fields here which require Drop!
+    // See the comment below in do_finish().
```
xzfc, line 506:
```suggestion
    // See the comment below in finish().
```

src/builder.rs
```diff
@@ -527,6 +529,15 @@ impl EntryWriter<'_> {
 
     /// Finish writing the current entry in the archive.
     pub fn finish(self) -> io::Result<()> {
+        // NOTE: This is an optimization for "fallible destructuring".
+        // we want finish() to return an error, but we also need to invoke
```
xzfc, line 533:
```suggestion
        // We want finish() to return an error, but we also need to invoke
```

xzfc (approved):

PR #382, opened 2024-10-21
Very basic Pax header writer

gwitrand-ovh:
Basic implem of PaxBuilder that can add the header with the data provided, implem on tar::builder

Today the lib is able to read the pax headers, even if they have arbitrary values but can't write some headers back. This update lets you do that very simply.

Usage:
```rust
let mut pax_builder: PaxBuilder = PaxBuilder::new();

let value: String = "a_value".to_string();
let key: String = "a_key".to_string();

pax_builder.add(&key, &value);

let res_write_pax: Result<(), io::Error> = my_open_tar_archive.append_pax_extensions(&pax_builder);
```

### DIFF
--- a/src/pax.rs
+++ b/src/pax.rs
@@ -1,5 +1,6 @@
 #![allow(dead_code)]
 use std::io;
+use std::io::Write;
 use std::slice;
 use std::str;
 
@@ -143,5 +144,50 @@ impl<'entry> PaxExtension<'entry> {
     /// Returns the underlying raw bytes for this value of this key/value pair.
     pub fn value_bytes(&self) -> &'entry [u8] {
         self.value
+    }
+}
+
+/// Extension trait for `Builder` to append PAX extended headers.
+impl<T: Write> crate::Builder<T> {
+    /// Append PAX extended headers to the archive.
+    ///
+    /// Takes in an iterator over the list of headers to add to convert it into a header set formatted.
+    ///
+    /// Returns io::Error if an error occurs, else it returns ()
+    pub fn append_pax_extensions<'a>(
+        &mut self,
+        headers: impl IntoIterator<Item = (String, String)>,
+    ) -> Result<(), io::Error> {
+        // Store the headers formatted before write
+        let mut data: Vec<u8> = Vec::new();
+
+        // For each key in headers, convert into a sized space and add it to data.
+        // This will then be written in the file
+        for (key, value) in headers {
+            let mut len_len = 1;
+            let mut max_len = 10;
+            let rest_len = 3 + key.len() + value.len();
+            while rest_len + len_len >= max_len {
+                len_len += 1;
+                max_len *= 10;
+            }
+            let len = rest_len + len_len;
+            write!(&mut data, "{} {}={}\n", len, key, value)?;
+        }
+
+        // Ignore the header append if it's empty.
+        if data.is_empty() {
+            return Ok(());
+        }
+
+        // Create a header of type XHeader, set the size to the length of the
+        // data, set the entry type to XHeader, and set the checksum
+        // then append the header and the data to the archive.
+        let mut header = crate::Header::new_ustar();
+        let data_as_bytes: &[u8] = &data;
+        header.set_size(data_as_bytes.len() as u64);
+        header.set_entry_type(crate::EntryType::XHeader);
+        header.set_cksum();
+        self.append(&header, data_as_bytes)
     }
 }
--- a/src/pax.rs
+++ b/src/pax.rs
@@ -154,9 +154,9 @@ impl<T: Write> crate::Builder<T> {
     /// Takes in an iterator over the list of headers to add to convert it into a header set formatted.
     ///
     /// Returns io::Error if an error occurs, else it returns ()
-    pub fn append_pax_extensions<'a>(
+    pub fn append_pax_extensions<'key, 'value>(
         &mut self,
-        headers: impl IntoIterator<Item = (String, String)>,
+        headers: impl IntoIterator<Item = (&'key str, &'value [u8])>,
     ) -> Result<(), io::Error> {
         // Store the headers formatted before write
         let mut data: Vec<u8> = Vec::new();
@@ -172,7 +172,9 @@ impl<T: Write> crate::Builder<T> {
                 max_len *= 10;
             }
             let len = rest_len + len_len;
-            write!(&mut data, "{} {}={}\n", len, key, value)?;
+            write!(&mut data, "{} {}=", len, key)?;
+            data.extend_from_slice(value);
+            data.push(b'\n');
         }
 
         // Ignore the header append if it's empty.
--- a/tests/entry.rs
+++ b/tests/entry.rs
@@ -1,7 +1,8 @@
 extern crate tar;
 extern crate tempfile;
 
-use std::fs::{create_dir, File};
+use std::fs::create_dir;
+use std::fs::File;
 use std::io::Read;
 
 use tempfile::Builder;
--- a/tests/all.rs
+++ b/tests/all.rs
@@ -932,12 +932,12 @@ fn pax_simple_write() {
     let file: File = t!(File::create(&pax_path));
     let mut ar: Builder<BufWriter<File>> = Builder::new(BufWriter::new(file));
 
-    let mut pax_builder: Vec<(String, String)> = Vec::new();
-    let key: String = "arbitrary_pax_key".to_string();
-    let value: String = "arbitrary_pax_value".to_string();
-    pax_builder.push((key, value));
+    let pax_extensions = [
+        ("arbitrary_pax_key", b"arbitrary_pax_value".as_slice()),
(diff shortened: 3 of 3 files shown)


src/pax.rs
```diff
@@ -145,3 +146,74 @@ impl<'entry> PaxExtension<'entry> {
         self.value
     }
 }
+
+/// A builder for PAX extended headers.
+/// It has the data for the extended header
+/// and a status on whether the extended header has been updated.
+pub struct PaxBuilder {
```
cgwalters, line 153: I don't quite understand what this buys us over just a `Vec<u8>`...the "updated" field is always equivalent to `!data.is_empty()` right?

src/pax.rs
```diff
@@ -145,3 +146,74 @@ impl<'entry> PaxExtension<'entry> {
...
+    }
+
+    /// Getter for the data field as bytes.
+    fn as_bytes(&self) -> &[u8] {
+        &self.data
+    }
+}
+
+/// Extension trait for `Builder` to append PAX extended headers.
+pub trait BuilderExt {
```
cgwalters, line 196: It looks like this code likely originated outside of this project as an extension trait which makes sense, but I don't see the need for us to have an extension trait for our own structures.

src/pax.rs
```diff
@@ -145,3 +146,74 @@ impl<'entry> PaxExtension<'entry> {
...
+    /// Append PAX extended headers to the archive.
+    fn append_pax_extensions(&mut self, headers: &PaxBuilder) -> Result<(), io::Error>;
+}
+
+/// Extension trait for `Builder` to append PAX extended headers.
+impl<T: Write> BuilderExt for crate::Builder<T> {
+    /// Append PAX extended headers to the archive.
+    fn append_pax_extensions(&mut self, headers: &PaxBuilder) -> Result<(), io::Error> {
```
cgwalters, line 204: I think we can drop `PaxBuilder` and have this API look like:

```
fn append_pax_extensions(&mut self, headers: impl IntoIterator<Item=(&[u8], &[u8])>) -> Result<(), io::Error>
```

or so?

Then we can do the creation of the temporary `Vec` inside this function.

cgwalters (commented):
Thanks for your contribution! Procedural notes:

- If you wouldn't mind please squash your commits, I don't think any of the 4 stand alone
- Can you rework your commit message to be "Linux kernel style" which I try to use, see e.g. https://cbea.ms/git-commit/

I'd propose something like this:

> Add support for writing PAX headers
>
> While this crate has always supported *reading* these attributes,
> writing them was up to the user. Add an API to do this.

cgwalters (changes requested):
Marking requested changes per above

src/pax.rs
gwitrand-ovh: Indeed, the updated on my first version was used for something else that I dropped.
Switching to check if empty

src/pax.rs
gwitrand-ovh: Switch to 
`fn append_pax_extensions<'a>(&mut self, headers: impl IntoIterator<Item=(String, String)>) -> Result<(), io::Error>` 

Didn't use [u8] because it was clashing with the write that I moved and &str had ownership / drop conflicts

gwitrand-ovh:
> Thanks for your contribution! Procedural notes:
> 
>     * If you wouldn't mind please squash your commits, I don't think any of the 4 stand alone
> 
>     * Can you rework your commit message to be "Linux kernel style" which I try to use, see e.g. https://cbea.ms/git-commit/
> 
> 
> I'd propose something like this:
> 
> > Add support for writing PAX headers
> > While this crate has always supported _reading_ these attributes,
> > writing them was up to the user. Add an API to do this.



> Marking requested changes per above

All comments fixed.
Also updated the tests and code doc to comply.

cgwalters:
Can you add these changes? First we don't need to require passing owned data, and according to the spec the value can be arbitrary binary data.

```
diff --git a/src/pax.rs b/src/pax.rs
index 5a06967..d149428 100644
--- a/src/pax.rs
+++ b/src/pax.rs
@@ -154,9 +154,9 @@ impl<T: Write> crate::Builder<T> {
     /// Takes in an iterator over the list of headers to add to convert it into a header set formatted.
     ///
     /// Returns io::Error if an error occurs, else it returns ()
-    pub fn append_pax_extensions<'a>(
+    pub fn append_pax_extensions<'key, 'value>(
         &mut self,
-        headers: impl IntoIterator<Item = (String, String)>,
+        headers: impl IntoIterator<Item = (&'key str, &'value [u8])>,
     ) -> Result<(), io::Error> {
         // Store the headers formatted before write
         let mut data: Vec<u8> = Vec::new();
@@ -172,7 +172,9 @@ impl<T: Write> crate::Builder<T> {
                 max_len *= 10;
             }
             let len = rest_len + len_len;
-            write!(&mut data, "{} {}={}\n", len, key, value)?;
+            write!(&mut data, "{} {}=", len, key)?;
+            data.extend_from_slice(value);
+            data.push(b'\n');
         }
 
         // Ignore the header append if it's empty.
diff --git a/tests/all.rs b/tests/all.rs
index 9f93de4..3caff4b 100644
--- a/tests/all.rs
+++ b/tests/all.rs
@@ -932,12 +932,12 @@ fn pax_simple_write() {
     let file: File = t!(File::create(&pax_path));
     let mut ar: Builder<BufWriter<File>> = Builder::new(BufWriter::new(file));
 
-    let mut pax_builder: Vec<(String, String)> = Vec::new();
-    let key: String = "arbitrary_pax_key".to_string();
-    let value: String = "arbitrary_pax_value".to_string();
-    pax_builder.push((key, value));
+    let pax_extensions = [
+        ("arbitrary_pax_key", b"arbitrary_pax_value".as_slice()),
+        ("SCHILY.xattr.security.selinux", b"foo_t"),
+    ];
 
-    t!(ar.append_pax_extensions(pax_builder.into_iter()));
+    t!(ar.append_pax_extensions(pax_extensions));
     t!(ar.append_file("test2", &mut t!(File::open(&pax_path))));
     t!(ar.finish());
     drop(ar);
@@ -950,9 +950,11 @@ fn pax_simple_write() {
     assert!(pax_headers.is_some(), "pax_headers is None");
     let mut pax_headers = pax_headers.unwrap();
     let pax_arbitrary = t!(pax_headers.next().unwrap());
-
     assert_eq!(pax_arbitrary.key(), Ok("arbitrary_pax_key"));
     assert_eq!(pax_arbitrary.value(), Ok("arbitrary_pax_value"));
+    let xattr = t!(pax_headers.next().unwrap());
+    assert_eq!(xattr.key().unwrap(), pax_extensions[1].0);
+    assert_eq!(xattr.value_bytes(), pax_extensions[1].1);
 
     assert!(entries.next().is_none());
 }

```

tests/entry.rs
```diff
@@ -1,7 +1,8 @@
 extern crate tar;
 extern crate tempfile;
 
-use std::fs::{create_dir, File};
+use std::fs::create_dir;
```
cgwalters, line 4: I don't care too much but this change seem spurious i.e. unnecessary

src/pax.rs
```diff
@@ -145,3 +146,50 @@ impl<'entry> PaxExtension<'entry> {
         self.value
     }
 }
+
+/// Extension trait for `Builder` to append PAX extended headers.
```
cgwalters, line 150: This isn't an extension trait anymore

cgwalters (approved):
Just two minor nits but they're not important so merging, we can do followup fixes if we want.

Thanks for your contribution!

cgwalters:
When I was playing with this I noticed that my suggestion of using `key: &str` clashes with the existing `PaxExtensions` reader, which gives back `&[u8]`. 

I dunno how much of a real world problem this is. It does sure look to me like in e.g. [the python tar parser](https://github.com/python/cpython/blob/74cf5967f345f944f1c308e5e30fee21b438eb6c/Lib/tarfile.py#L1498) it's actually expected to be UTF-8.

cgwalters:
For general interest I used this to fix a bug in my software - thank you for submitting this patch! cc https://github.com/ostreedev/ostree-rs-ext/pull/679

tests/entry.rs
gwitrand-ovh: Didn't realize at the time, not intended. Won't fix since it's not that important

gwitrand-ovh:
> When I was playing with this I noticed that my suggestion of using `key: &str` clashes with the existing `PaxExtensions` reader, which gives back `&[u8]`.
> 
> I dunno how much of a real world problem this is. It does sure look to me like in e.g. [the python tar parser](https://github.com/python/cpython/blob/74cf5967f345f944f1c308e5e30fee21b438eb6c/Lib/tarfile.py#L1498) it's actually expected to be UTF-8.

This could be fixed, but I'm not sure it would add any "bonus" as str is used for an utf8 string for the key anyway

gwitrand-ovh:
Made a PR if want to switch the usage of key but it might be a breaking change if already used by someone

https://github.com/alexcrichton/tar-rs/pull/391/files

PR #385, opened 2024-10-23
OSS-Fuzz: OSS-Fuzz fuzzing integration

arthurscchan:
Hi! Would you be interested in setting up fuzzing for the tar-rs module via OSS-Fuzz?

Fuzzing is essentially a stress-testing approach used to find bugs in software, and OSS-Fuzz is a free service run by Google that continuously fuzzes important open-source projects. Integrating your module with OSS-Fuzz could help uncover memory corruption issues that may exist.

This PR adds a Cargo fuzz configuration along with 3 fuzzers for the tar-rs module. In combination with an initial attempt in OSS-Fuzz (https://github.com/google/oss-fuzz/pull/12645), it enables OSS-Fuzz to fuzz the tar-rs module while keeping the fuzzers upstream for further modification and expansion. If you're happy to proceed with the integration and store the fuzzers upstream, please let me know, and I'd be glad to provide more details if needed.

The only thing required at this point is an email associated with a Google account, which will be used to receive notifications when bugs are found.


fuzz/fuzz_targets/archive.rs
```diff
@@ -0,0 +1,139 @@
...
+        let mut tar_data = Vec::new();
+        let _ = tar_file.read_to_end(&mut tar_data);
+        let mut tar_cursor = Cursor::new(tar_data);
+        let mut archive = Archive::new(&mut tar_cursor);
+        let _ = archive.unpack(temp_dir.path()).ok();
+    }
+
+    // Fuzz archive and builder
+    for i in 0..3 {
+        let name_data = &data[i * 5 % data.len()..(i * 5 + 5) % data.len()];
```
cgwalters, line 118: What is this doing? A comment would be useful

fuzz/fuzz_targets/builder.rs
```diff
@@ -0,0 +1,67 @@
...
+use tar::Builder;
+use tempfile::{tempdir, tempfile};
+
+fuzz_target!(|data: &[u8]| {
+    // Initialization
+    let random_bool = data.first().map(|&b| b % 2 == 0).unwrap_or(false);
+    let temp_dir = tempdir().expect("");
+    
+    // Create a temporary file for testing
+    if let Ok(temp_file) = tempfile() {
```
cgwalters, line 29: Why would we ignore errors here?

fuzz/fuzz_targets/archive.rs
```diff
@@ -0,0 +1,139 @@
...
+use std::fs::{File, OpenOptions};
+use std::io::{Cursor, Write, Read};
+use tar::{Archive, Builder, EntryType, Header};
+use tempfile::tempdir;
+use std::convert::TryInto;
+use std::str;
+
+fuzz_target!(|data: &[u8]| {
+    // Skip this iteration when data is not enough
+    if data.len() < 10 {
```
cgwalters, line 28: How about a 
`let minbuf_size = 10;` above this and then use it below

fuzz/fuzz_targets/archive.rs
```diff
@@ -0,0 +1,139 @@
...
+        return;
+    }
+
+    // Create temp file and dir
+    let temp_dir = tempdir().unwrap();
+    let file_name = match str::from_utf8(&data[0..data.len().min(10)]) {
```
cgwalters, line 34: There's some "manual array splitting" here that looks not-wrong but I think would be cleaner if we did it like:

```
let (file_name, data) = data.split_at(minbuf_size);
let file_name = str::from_utf8(file_name)...
```

fuzz/fuzz_targets/archive.rs
```diff
@@ -0,0 +1,139 @@
...
+        Ok(name) => name.to_string(),
+        Err(_) => "default_file_name".to_string(),
+    };
+    let dir_name = match str::from_utf8(&data[data.len().min(10)..data.len().min(20)]) {
```
cgwalters, line 38: Then from here we can just reference `data` as if it was consumed.

Another way to do all of this is just to wrap `data` in a https://doc.rust-lang.org/std/io/struct.Cursor.html and call `read` or `read_exact` etc on it instead.

fuzz/fuzz_targets/archive.rs
```diff
@@ -0,0 +1,139 @@
...
+        1 => EntryType::Directory,
+        2 => EntryType::Symlink,
+        3 => EntryType::hard_link(),
+        _ => EntryType::character_special(),
+    };
+
+    // Initilaise header
+    let mut header = Header::new_gnu();
+    let file_size = u64::from_le_bytes(
+        data.get(1..9)
```
cgwalters, line 61: We're reading the same data for the filename, right? That's fine, but shouldn't we try to use independent sources for this? IOW let's just consistently read some bytes from data at each step instead of slicing it.

cgwalters (commented):
This is great, thanks! I only took a start of a review on this but I love the idea.

cgwalters:
> The only thing required at this point is an email associated with a Google account, which will be used to receive notifications when bugs are found.

My gmail is cgwalters@gmail.com, the other most active maintainers are @xzfc and @alexcrichton so let's see if they have and want to provide gmail accounts too.

(But that said...why can't oss-fuzz learn to e.g. file security issues on GH?)

fuzz/fuzz_targets/builder.rs
arthurscchan: It is because the logic only meaningful once the temp_file does successfully created. If the tempfile creation is failed, it is not worth to continue and simply should skip to the next iteration.

fuzz/fuzz_targets/archive.rs
arthurscchan: I will fix it.

fuzz/fuzz_targets/builder.rs
cgwalters: We'd only fail to create a temporary file in the same situations in which we fail to run `tempdir`.

And why are we making both a tempdir and a tempfile?

Also tangential but `expect("")` is pointless, just use `unwrap()`.

I think we should just consistently `unwrap()` here.

cgwalters (changes requested):
Marking changes requested to track review state above

fuzz/fuzz_targets/archive.rs
arthurscchan: Have fixed both dir_name and file_name

fuzz/fuzz_targets/archive.rs
arthurscchan: Have fixed both dir_name and file_name

fuzz/fuzz_targets/archive.rs
arthurscchan: Yes, I have fixed them.

fuzz/fuzz_targets/archive.rs
arthurscchan: Sorry, this section should be removed. It is some leftover code. Apology for that.

alexcrichton:
Personally I'm all for more fuzzing. I'm [cc'd on the wasmtime project](https://github.com/google/oss-fuzz/blob/bf3ff619b18d720c3c9dc4aa1523819f6fc1f9d1/projects/wasmtime/project.yaml#L5) on oss-fuzz and we use it to great benefit there. In that sense 👍  from me.

On the specifics of the fuzzers here one thing I might recommend is to more heavily use the `arbitrary` crate to make parsing the input `data` easier. I'd also be a bit worried about the coverage of these fuzzers, for example the `tar.rs` fuzzer looks more like a unit test than a fuzzer since the data is just written to the filesystem. For `builder.rs` that also looks more like a unit tests since the set of things that can happen in the fuzzer is so small (more-or-less only looking at the first byte). The `archive.rs` fuzzer looks like it's got some interesting bits going on but it's a bit hard to read/follow I find and I think that `arbitrary` would help clean it up.

Overall though I think the idea of fuzzing is good, but I think it might be best to shore up what's being fuzzed here. For example this might be a good candidate for differential fuzzing against a different `tar` library or similar (e.g. linking to libtar in C if that's reasonably easy to do). Either that or perhaps fuzzing things like assigning arbitrary filenames and ensuring that no files are created outside of the folder during extraction (stuff like that). I'll note though that in general fuzzers are most effective when they do no I/O since that enables them to go much faster.

arthurscchan:
@cgwalters I have updated the fuzzers to include more randomness and clean up with arbitrary crate. I have also removed builder.rs which have duplicated targets from the other two fuzzers to make it more clear.

arthurscchan:
@alexcrichton Thank you for your suggestions. I have updated the fuzzers accordingly. I have removed `builder.rs` as it contains much of the duplicated logic from the other two fuzzers. I have also included more randomness in file structure and contents using the `arbitrary` crate for both `tar.rs` and `archive.rs`. The aim of these fuzzers is to explore and test the code for tar archiving and extraction functionalities. I believe the fuzzers are much improved now, both in terms of randomness and targeted functions. Please do share any further comments on improving the fuzzing targets or approaches to enhance them. Thanks again for your suggestions and feedback.

fuzz/fuzz_targets/archive.rs
```diff
@@ -0,0 +1,133 @@
...
+                // Handle other types with appropriate mock content or skip unsupported
+                let mut cursor = Cursor::new(entry.content.clone());
+                if builder.append_data(&mut header, entry.path.as_str(), &mut cursor).is_err() {
+                    continue;
+                }
+            }
+        }
+    }
+
+    if let Ok(mut temp_file) = File::create(&temp_file_path) {
```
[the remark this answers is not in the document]
arthurscchan: Yes, unwrap here is much better handling. Thanks for the suggestion. I have fixed it.

fuzz/fuzz_targets/archive.rs
```diff
@@ -0,0 +1,133 @@
...
+        header.set_entry_type(entry_type);
+
+        // Process entry types based on directory structure and content
+        match entry_type {
+            EntryType::Directory => {
+                let dir_path = temp_dir.path().join(&entry.path);
+                if std::fs::create_dir_all(&dir_path).is_err() {
+                    continue;
+                }
+                if builder.append_dir(entry.path.clone(), &dir_path).is_err() {
```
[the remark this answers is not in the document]
arthurscchan: Sorry for overlook the possible directory traversal problem. I have revamp the logic a bit to include checking and sanitising to ensure it does not happened. And the joining of temp_dir is done in earlier logic.

cgwalters:
Thanks for your work on this! Your filtering of the pathnames looks sane, but I still worry about having code that reads or writes to the filesystem as a result of the fuzzer. In CI, it doesn't really matter - I'm *sure* that the oss-fuzz runners are sandboxed heavily.

But someone trying to reproduce a fuzzing failure locally is probably not by default, and it'd be pretty unfortunate if we somehow ended up writing into their `$HOME` or so for example.

The safest fix is really to avoid all filesystem APIs. I should have recommended this earlier but we could also use https://crates.io/crates/cap-std which itself acts as a sandbox.

Alternatively - and this may be the easiest - we could have the fuzz targets bomb out like this:

```
if !std::env::var_os("TAR_FUZZ_ACKNOWLEDGE_SANDBOX_RECOMMENDED").is_some() {
  anyhow::bail!("These fuzzing targets may read/write to the filesystem; please run in a sandbox and set TAR_FUZZ_ACKNOWLEDGE_SANDBOX_RECOMMENDED");
```
or so?

NobodyXu:
> The safest fix is really to avoid all filesystem APIs. I should have recommended this earlier but we could also use https://crates.io/crates/cap-std which itself acts as a sandbox.

Just my 2c, but it seems that having a sans-io interface would help fuzzing, as it ensures no I/O?

fuzz/fuzz_targets/archive.rs
```diff
@@ -0,0 +1,144 @@
...
+// Implement Arbitrary for ArchiveEntry
+impl<'a> Arbitrary<'a> for ArchiveEntry {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let path: String = u.arbitrary::<&str>()?.to_string();
+        let entry_type: u8 = u.arbitrary()?;
+        let content: Vec<u8> = u.arbitrary()?;
+        
+        Ok(ArchiveEntry { path, entry_type, content })
+    }
+}
```
alexcrichton, line 42: This is relatively simple so I might also recommend `derive(Arbitrary)` which can be enabled through the `derive` feature of the `arbitrary` crate

fuzz/fuzz_targets/tar.rs
```diff
@@ -0,0 +1,137 @@
...
+// Implement Arbitrary for FuzzInput
+impl<'a> Arbitrary<'a> for FuzzInput {
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        Ok(FuzzInput {
+            data: u.arbitrary()?,
+            file_name: u.arbitrary::<&str>()?.to_string(),
+            link_path: u.arbitrary::<&str>()?.to_string(),
+            target_path: u.arbitrary::<&str>()?.to_string(),
+            entry_type: u.arbitrary()?,
+            metadata_size: u.arbitrary::<u64>()? % 1000, // Limit size to a reasonable max
```
alexcrichton, line 43: For this I might recommend `u.int_in_range(...)`

alexcrichton (commented):
Fuzzers look good to me, thanks! I might second the cap-std suggestion for perhaps creating the directory to crate an archive from. That can help remove the tests for safe paths I think since it should automatically deny access outside the directory (if I understand it right)

While skipping I/O entirely would be best I do realize that `tar` is intimately tied with the filesystem most of the time so it might just not be possible to skip it for these fuzzers.

fuzz/fuzz_targets/archive.rs
arthurscchan: Thanks for the suggestion, I have added the derive_arbitrary crate to handle and simplify that.

fuzz/fuzz_targets/tar.rs
arthurscchan: I have fixed that.

arthurscchan:
@alexcrichton @cgwalters I have revamped the logic, using derive_abitrary crate and change the size check to use int_in_range. I have also implemented the sandbox_dir with cap_std and remove unnecessary path checking and sanitisation (since cap_std won't allow going outside from the sandbox_dir). I agree that it is a better handling of the I/O process. I actually second the idea that I/O is kinda not avoidable in fuzzing this project since tar is coupled with I/O process in some sence, and that is also one of the worthy fuzzing target in the project.

fuzz/fuzz_targets/archive.rs
```diff
@@ -0,0 +1,124 @@
...
+        Ok(val) => val,
+        Err(_) => return,
+    };
+
+    // Create a sandbox directory with cap_std
+    let temp_dir = match tempdir() {
+        Ok(dir) => dir,
+        Err(_) => return,
+    };
+    let sandbox_dir = match Dir::open_ambient_dir(temp_dir.path(), ambient_authority()) {
```
cgwalters, line 53: Fine as is though there's also `cap_tempfile` which streamlines this

cgwalters (approved):
Thanks so much for your work on this! I think we can get this in and see if fuzzing at scale turns up anything.

Am I correct in that basically what this fuzzing coverage should find is an unexpected `panic` or so? We're swallowing/ignoring most `Err`.

I think it'd be quite interesting to have a followup for Alex's suggestion of differential fuzzing, e.g. something like validate that a tar we create can "round trip" especially through other tar implementations like GNU tar, libarchive, the Go `encoding/tar` etc.
That would likely turn up a lot of edge cases in interoperability.

arthurscchan:
@cgwalters Yes, we are aiming to find unexpected panic, or even deadly signal like Segmentation Fault or else. Thanks for merging the fuzzers in. I will go on and add you and the other two maintainer's email to the OSS-Fuzz integration and it should be ready to go.

PR #387, opened 2024-11-08
sparse test: fix for 64k page size

Fabian-Gruenbichler:
this test failed on Debian's CI on ppc64el, because the page size is 64k there, which means the block size on tmpfs is as well, so attempting to create 32k big holes cannot work.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1114,8 +1114,9 @@ impl<W: Write> Drop for Builder<W> {
 mod tests {
     use super::*;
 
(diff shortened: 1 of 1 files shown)


Fabian-Gruenbichler:
alternatively, the block size here could be `#[cfg()]`-ed based on the target arch

Fabian-Gruenbichler:
link to the failing CI test: https://ci.debian.net/packages/r/rust-tar/testing/ppc64el/54192787/#S9

alexcrichton (approved):

Issue #369, opened 2024-07-03
Cannot `tar::Archive.unpack()` an archive with long link in it


schneems:
## Context 

I'm using `tar` to extract the contents downloaded from https://repo1.maven.org/maven2/org/jruby/jruby-dist/9.4.8.0/jruby-dist-9.4.8.0-bin.tar.gz for re-packaging it and uploading it to S3 https://github.com/heroku/docker-heroku-ruby-builder/blob/9e64b4401be4df7c158c9253290f6a3248927023/shared/src/lib.rs#L24-L31. 

Unfortunately I've learned that this file uses ././@LongLink and it seems that the rust tar archive truncates the file by default. To demonstrate the issue I made a reproduction https://github.com/schneems/tar_long_link_repro.

## Reproduction

```
$ git clone https://github.com/schneems/tar_long_link_repro
$ cd tar_long_link_repro
```

Then run it:

```
$ cargo run
```

## Expected

I expect that `tmp/rust-extracted/jruby-9.4.8.0/lib/ruby/stdlib/bundler/vendor/molinillo/lib/molinillo/delegates` will contain a file `specification_provider.rb`.

## Actual

It does not, the command fails an assertion:

```
$ cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/tar_long_link_repro`
thread 'main' panicked at src/main.rs:32:5:
expected ["resolution_state.rb", "specification_provide"] to include "specification_provider.rb" but it did not
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Note the filename seems truncated:

```
$ ls tmp/rust-extracted/jruby-9.4.8.0/lib/ruby/stdlib/bundler/vendor/molinillo/lib/molinillo/delegates
resolution_state.rb	specification_provide
```

## More

I see there that the header and entry inside of tar are aware of the concept of long names (https://docs.rs/tar/0.4.41/tar/struct.Header.html?search=long), but I'm not sure how to unpack a Gnu tar file that has them. If the feature exists, we could possibly update the documentation to make it clearer.

schneems:
I might have misdiagnosed the problem. Using bsdtar, it doesn't look like the archive has a long link in it:

```
$ tar -tvf fixtures/jruby-dist-9.4.8.0-bin.tar.gz | grep LongLink ; echo "done"
done
```

But for some reason `tar-rs` thinks it does.

headius:
[...]

PR #389, opened 2024-11-12
Fix long names (GNUTYPE_LONGNAME)

frederikbosch:
The patch suggested by @RazerM in #369 is fixing my problem of a truncated file name during extraction. In my case the long file names ended up in the root of the extraction directory rather than the sub-directory they have in the archive.

>  I'm also not sure what the best way forward is for handling this is. I can't find anything about why these bytes might be different than expected, if not just a mixup of \0 and 0.

Is what @RazerM said last week. I'll be honest here: I don't have a clue at all. I'm just creating a PR for the solution that is solving my problem.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -193,7 +193,8 @@ impl Header {
 
     fn is_ustar(&self) -> bool {
         let ustar = unsafe { cast::<_, UstarHeader>(self) };
-        ustar.magic[..] == b"ustar\0"[..] && ustar.version[..] == b"00"[..]
+        ustar.magic[..] == b"ustar\0"[..]
+             && (ustar.version[..] == b"00"[..] || ustar.version[..] == b"\0\0"[..])
     }
 
     fn is_gnu(&self) -> bool {
--- a/src/header.rs
+++ b/src/header.rs
@@ -194,7 +194,7 @@ impl Header {
     fn is_ustar(&self) -> bool {
         let ustar = unsafe { cast::<_, UstarHeader>(self) };
         ustar.magic[..] == b"ustar\0"[..]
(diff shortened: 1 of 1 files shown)


RazerM:
To be clear, the issue is not related to long links.

P.S. I did not submit this as a PR myself because it should probably not be part of `is_ustar` – the header in the problematic tar files is non-standard.

frederikbosch:
Are you sure about that @RazerM? In my case, I am creating archives in exactly the same manner. The only difference is the length of the added paths. I had a look in the library I created a long time ago. It contains [a specific `././@LongLink` case for files with dirnames longer than 155 bytes or filenames longer than 100 bytes](https://github.com/genkgo/archive-stream/blob/master/src/TarReader.php#L62).

RazerM:
The long links are stored in a header whose version bytes are "wrong", tar-rs does not look in invalid headers.

> I am creating archives

Using what? Sounds like it's creating out of spec headers.

frederikbosch:
Apparently, I am using `GNUTYPE_LONGNAME`, see the link to my archive generation code in the message above, which is [supported by Python](https://docs.python.org/3/library/tarfile.html#tarfile.GNUTYPE_LONGNAME).

RazerM:
I mean what program or library are you using to create the archive

frederikbosch:
See (this library)[https://github.com/genkgo/archive-stream/blob/master/src/TarReader.php#L62], the link references the specific line that is using the GNUTYPE_LONGNAME (L type header).

frederikbosch:
Maybe you are right, this is not about longname.

frederikbosch:
Thanks for waking me up. I have fixed the ustar version issue in my own library. Much appreciated.

PR #390, opened 2024-11-12
fix for archiving long paths that have path components starting with ".." crossing the 100-character mark

mitrandir77:
The gnu tar supports arbirary path length by putting path truncated to standard 100 chars into the header and the rest is appended to contents. `tar-rs` validates that no path components should be exactly  ".." but in this case when a component starting with ".." (for example file named "..some_file") gets truncated after 2 characters we hit this validation and can't tar such file.

I have verified that gnu `tar` command can handle such paths and actually puts truncated `..` in the header. This pull request makes `tar-rs` behave the same.

See tests for repro of the issue.

(I think the code for handling this special case is quite ugly - I'd appreciate suggestions for making it better)

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -768,7 +768,7 @@ fn prepare_header_path(dst: &mut dyn Write, header: &mut Header, path: &Path) ->
             Ok(s) => s,
             Err(e) => str::from_utf8(&data[..e.valid_up_to()]).unwrap(),
         };
-        header.set_path(truncated)?;
+        header.set_truncated_path_for_gnu_header(&truncated)?;
     }
     Ok(())
 }
--- a/src/header.rs
+++ b/src/header.rs
@@ -380,14 +380,21 @@ impl Header {
     /// use `Builder` methods to insert a long-name extension at the same time
     /// as the file content.
     pub fn set_path<P: AsRef<Path>>(&mut self, p: P) -> io::Result<()> {
-        self._set_path(p.as_ref())
+        self._set_path(p.as_ref(), false)
     }
 
-    fn _set_path(&mut self, path: &Path) -> io::Result<()> {
+    // Sets the truncated path for GNU header
+    //
+    // Same as set_path but skips some validations.
+    pub(crate) fn set_truncated_path_for_gnu_header<P: AsRef<Path>>(&mut self, p: P) -> io::Result<()> {
+        self._set_path(p.as_ref(), true)
+    }
+
+    fn _set_path(&mut self, path: &Path, is_truncated_gnu_long_path: bool) -> io::Result<()> {
         if let Some(ustar) = self.as_ustar_mut() {
             return ustar.set_path(path);
         }
-        copy_path_into(&mut self.as_old_mut().name, path, false).map_err(|err| {
+        copy_path_into(&mut self.as_old_mut().name, path, false, is_truncated_gnu_long_path).map_err(|err| {
             io::Error::new(
                 err.kind(),
                 format!("{} when setting path for {}", err, self.path_lossy()),
@@ -439,7 +446,7 @@ impl Header {
     }
 
     fn _set_link_name(&mut self, path: &Path) -> io::Result<()> {
-        copy_path_into(&mut self.as_old_mut().linkname, path, true).map_err(|err| {
+        copy_path_into(&mut self.as_old_mut().linkname, path, true, false).map_err(|err| {
             io::Error::new(
                 err.kind(),
                 format!("{} when setting link name for {}", err, self.path_lossy()),
@@ -991,7 +998,7 @@ impl UstarHeader {
         let bytes = path2bytes(path)?;
         let (maxnamelen, maxprefixlen) = (self.name.len(), self.prefix.len());
         if bytes.len() <= maxnamelen {
-            copy_path_into(&mut self.name, path, false).map_err(|err| {
+            copy_path_into(&mut self.name, path, false, false).map_err(|err| {
                 io::Error::new(
                     err.kind(),
                     format!("{} when setting path for {}", err, self.path_lossy()),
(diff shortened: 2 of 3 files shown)


src/header.rs
```diff
@@ -1540,17 +1556,29 @@ fn copy_into(slot: &mut [u8], bytes: &[u8]) -> io::Result<()> {
...
             }
             (Component::ParentDir, false) => {
-                return Err(other("paths in archives must not have `..`"));
+                if is_truncated_gnu_long_path && iter.peek().is_none() {
+                    // If it's last component of a gnu long path we know that there might be more
+                    // to the component than .. (the rest is stored elsewhere)
+                    {}
+                } else {
+                    return Err(other("paths in archives must not have `..`"));
+                }
```
cgwalters, line 1581: Why not
```
if is_long_path && iter.peek.is_some() {
    return Err(other("paths in archives must not have `..`"));
```
?

tests/all.rs
```diff
@@ -203,6 +203,36 @@ fn large_filename() {
...
+
+    let mut header = Header::new_gnu();
+    header.set_cksum();
+    header.set_mode(0o644);
+    header.set_size(4);
+
+    let mut long_name_with_dot_dot = "tdir/".repeat(19);
+    long_name_with_dot_dot.push_str("tt/..file");
+
+    t!(ar.append_data(&mut header, &long_name_with_dot_dot, &b"test"[..]));
```
cgwalters, line 221: I would use `.as_slice()` instead of the `&` and `[..]` but I see it's not new in the code.

src/header.rs
```diff
@@ -1540,17 +1556,29 @@ fn copy_into(slot: &mut [u8], bytes: &[u8]) -> io::Result<()> {
 /// * a nul byte was found
 /// * an invalid path component is encountered (e.g. a root path or parent dir)
 /// * the path itself is empty
-fn copy_path_into(mut slot: &mut [u8], path: &Path, is_link_name: bool) -> io::Result<()> {
+fn copy_path_into(
+    mut slot: &mut [u8],
+    path: &Path,
+    is_link_name: bool,
+    is_truncated_gnu_long_path: bool,
```
cgwalters, line 1563: This is OK as is but...I think functions with more than one `bool` parameter can be hard to read on the call side. From a quick glance they seem to be mutually exclusive? We could add a `copy_path_into_inner` that takes both bools, but have a thin wrapper like

```
fn copy_path_into_gnu_long(...) {
    copy_path_into_impl(...)
}
```
or so?

tests/all.rs
```diff
@@ -203,6 +203,41 @@ fn large_filename() {
...
 
+// This test checks very particular scenario where a path component starting
+// with ".." of a long path gets split at 100-byte mark so that ".." part goes
+// into header and gets interpreted as parent dir (and rejected) .
+#[test]
+fn large_filename_with_dot_dot_at_100_byte_mark() {
+    let mut ar = Builder::new(Vec::new());
+
+    let mut header = Header::new_gnu();
+    header.set_cksum();
```
cgwalters, line 214: calling `set_cksum` I think should always be the end thing, but here you can probably just drop it.

src/header.rs
```diff
@@ -1540,17 +1547,25 @@ fn copy_into(slot: &mut [u8], bytes: &[u8]) -> io::Result<()> {
...
         match (component, is_link_name) {
             (Component::Prefix(..), false) | (Component::RootDir, false) => {
                 return Err(other("paths in archives must be relative"));
             }
             (Component::ParentDir, false) => {
-                return Err(other("paths in archives must not have `..`"));
+                if is_truncated_gnu_long_path && iter.peek().is_none() {
+                    // If it's last component of a gnu long path we know that there might be more
+                    // to the component than .. (the rest is stored elsewhere)
+                    {}
```
cgwalters, line 1578: The explicit `{}` here is odd, how about inverting the condition:
```
if !(is_truncated && iter.peek.is_none()) {
    return Err(other("paths in archives must not have `..`"));
}
```

cgwalters (commented):
Only a superficial code style skim so far...I haven't deeply analyzed the code but it looks plausible

src/header.rs
mitrandir77: I'll change to:

``
if !is_long_path || iter.peek.is_some() {
``

src/header.rs
mitrandir77: This will make it easier to use the vanilla `copy_path_into` without considering the need for this setting - but I understand that in most cases that's the right choice so let's do it.
If it becomes a problem perhaps a better way would be to pass some option struct containing all the `bools` with settings.

mitrandir77:
@cgwalters Thanks for all the comments. I have fixed the coding style issues you raised making the code hopefully easier to review for the next round.

cgwalters (approved):

src/builder.rs
```diff
@@ -768,7 +768,7 @@ fn prepare_header_path(dst: &mut dyn Write, header: &mut Header, path: &Path) ->
             Ok(s) => s,
             Err(e) => str::from_utf8(&data[..e.valid_up_to()]).unwrap(),
         };
-        header.set_path(truncated)?;
+        header.set_truncated_path_for_gnu_header(&truncated)?;
```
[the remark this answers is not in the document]
mitrandir77: > but what I feel would improve things here is if we did a cast `as_gnu_mut()` and operated on that instead or so

I agree that it's a good idea but unfortunately a major refactor. From what I can see the problem starts at: [`prepare_header_path`](https://github.com/alexcrichton/tar-rs/blob/main/src/builder.rs#L744) function which acts as de-facto `set_path` that can handle GNU header. I think it's done this way because unlike other `set_path` functions (the vanilla and ustar one), this one doesn't just influence header structure but emits a new entry into archive containing that long path.

PR #391, opened 2024-11-25
Pax | Switching key to u8 and update on Pax header doc header

gwitrand-ovh:
Following PR #382 comments


gwitrand-ovh:
This might not be recommended since key() converts the key into str.

Writing the key through str might me more straight forward in the long run
image

PR #394, opened 2024-12-09
Replace hard-coded 512 with BLOCK_SIZE

xzfc:
Extracted from #298. Unlike the original PR, I made the new constant `u64` instead of `usize` to reduce amount of type casts.


cgwalters (approved):
Nice!

PR #395, opened 2025-01-12
use generics instead of dynamic dispatch internally in `tar::Builder`

GabrielDertoni:
**TL;DR** Using generics instead of dynamic dispatch allows for specialization and a performance win on modern Linux systems.

| bench                          | old time       | new time        | delta   |
|:-------------------------------|:---------------|:----------------|--------:|
| cache=yes size=583M files=23k  | 1.9 s ± 0.1 s  | 1.7 s ± 0.1 s   | -10.08% |
| cache=no  size=583M files=23k  | 7.3 s ± 0.2 s  | 6.7 s ± 0.3 s   |  -8.95% |
| cache=yes size=284M files=1k   | 500 ms ±  4 ms | 362 ms ± 28 ms  | -38.12% |
| cache=no  size=284M files=1k   | 918 ms ± 25 ms | 889 ms ± 29 ms  |  -3.26% |

(look [here](https://gist.github.com/GabrielDertoni/ac6832046e6eabec43d4e45da5a3b027) for full benchmark details)

This PR is specifically targeted to allow triggering the specialization of the `Write` implementations when using `tar::Builder`. In particular, the [stdlib has a specialization on Linux for `std::io::copy`](https://github.com/rust-lang/rust/blob/7bb98889534d597911263d647424557b414770ec/library/std/src/sys/pal/unix/kernel_copy.rs#L74) that allows it to use the [`copy_file_range` syscall](https://man7.org/linux/man-pages/man2/copy_file_range.2.html). Unfortunately the code as-is is not capable of performing such specialization (or potentially others) since it relies on dynamic dispatch internally. This PR changes the underlying implementation to uses generics instead of dynamic dispatch. This understandably comes at the potential cost of increased code size but I'd argue this choice is already exposed to users that should be willing to pay the code size cost when using `tar::Builder<T>` directly instead of `tar::Builder<dyn Write>`. Users looking for reduced code footprint can explicitly opt in through using `tar::Builder<dyn Write>` instead.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -583,14 +583,18 @@ impl Drop for EntryWriter<'_> {
     }
 }
 
-fn append(mut dst: &mut dyn Write, header: &Header, mut data: &mut dyn Read) -> io::Result<()> {
+fn append<W: Write + ?Sized>(
+    mut dst: &mut W,
+    header: &Header,
+    mut data: &mut dyn Read,
+) -> io::Result<()> {
     dst.write_all(header.as_bytes())?;
     let len = io::copy(&mut data, &mut dst)?;
     pad_zeroes(&mut dst, len)?;
     Ok(())
 }
 
-fn pad_zeroes(dst: &mut dyn Write, len: u64) -> io::Result<()> {
+fn pad_zeroes<W: Write + ?Sized>(dst: &mut W, len: u64) -> io::Result<()> {
     let buf = [0; BLOCK_SIZE as usize];
     let remaining = BLOCK_SIZE - (len % BLOCK_SIZE);
     if remaining < BLOCK_SIZE {
@@ -599,8 +603,8 @@ fn pad_zeroes(dst: &mut dyn Write, len: u64) -> io::Result<()> {
     Ok(())
 }
 
-fn append_path_with_name(
-    dst: &mut dyn Write,
+fn append_path_with_name<W: Write + ?Sized>(
+    dst: &mut W,
     path: &Path,
     name: Option<&Path>,
     options: BuilderOptions,
@@ -641,8 +645,8 @@ fn append_path_with_name(
 }
 
 #[cfg(unix)]
-fn append_special(
-    dst: &mut dyn Write,
+fn append_special<W: Write + ?Sized>(
+    dst: &mut W,
     path: &Path,
     stat: &fs::Metadata,
     mode: HeaderMode,
@@ -684,8 +688,8 @@ fn append_special(
     Ok(())
 }
 
-fn append_file(
-    dst: &mut dyn Write,
+fn append_file<W: Write + ?Sized>(
+    dst: &mut W,
     path: &Path,
     file: &mut fs::File,
     options: BuilderOptions,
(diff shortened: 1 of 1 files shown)


cgwalters:
> In particular, the [stdlib has a specialization on Linux for std::io::copy](https://github.com/rust-lang/rust/blob/7bb98889534d597911263d647424557b414770ec/library/std/src/sys/pal/unix/kernel_copy.rs#L74) that allows it to use the [copy_file_range syscall](https://man7.org/linux/man-pages/man2/copy_file_range.2.html).

And that would happen here when your tar output is to a file, and you're providing a `File` instance as input e.g. `append()`? OK. (Or a pipe/socket for either of those)

OK, though I do kind of question how often this occurs, because *most* use cases for tar end up being compressed (i.e. the output side is not a fd).

Can you at briefly explain your higher level use case and what led you to try this optimization?

GabrielDertoni:
Fair point on the fact that this is mostly used for compression purposes. However, this specialization also applies to writers  like `BufWriter<T>` which can have their larger internal buffers used directly, avoiding one unnecessary copy.

I stumbled on this when someone told me they were trying to archive a python environment and taring was taking a significant portion of the time. I then looked through the implementation and realized many syscalls were being issued that could be significantly reduced. Although this optimization may dramatically reduce the number of syscalls in some cases it seldom results in as big of a time reduction, but it is not negligible.

This can be quite useful whey you'd like to write to an archive temporarily, even if it is going to be compressed afterwards, or you want to pipe that result into another program to handle the compression. I guess what surprised me is that since `tar::Builder` is a generic type my performance expectation is that it will monorphize and benefit where it can from that so I found it quite odd that internally it uses dynamic dispatch.

cgwalters:
There are certainly use cases for uncompressed tar; over in OCI land we were discussing that uncompressed tar should used when the data embedded is already compressed - AI models can be like this. This helps avoid a problem in OCI because it checksums compressed data, and that can make it hard to ensure reproducible caching.

Anyways though...the change sounds reasonable but, this crate is *widely* used and I personally try to be pretty conservative with what we merge. This type of change has the potential to cause regressions (probably in code size mostly - but certainly not exclusively).

Can you please provide a bit more benchmarking in a targeted use case?

> I stumbled on this when someone told me they were trying to archive a python environment and taring was taking a significant portion of the time. I then looked through the implementation and realized many syscalls were being issued that could be significantly reduced. Although this optimization may dramatically reduce the number of syscalls in some cases it seldom results in as big time reduction, but it is not negligible.

A bit more data? Are we talking 1%? 10%? As far as syscall traffic, of course others have been looking at io-uring based work (see https://github.com/alexcrichton/tar-rs/pull/355 ), but that's a much deeper change.

GabrielDertoni:
Hi, I understand you being conservative and I appreciate it! Unfortunately I am not able to provide benchmarks more specific than I have already provided in the original description, but the perf benefits are in the order showed on the previous benchmarks (up to 38%, but 10% mostly). I am certainly looking at io-uring based alternatives myself but I agree that's a much larger change and may not fit in this crate. This PR though I thought would benefit upstream for a sensible optimization without too many downsides. And I'd be interested to know what other downsides you think this would cause downstream and if there is some specific example you might be worried about. I'd also be happy to consider feature-flagging this if you think that'd be more appropriate.

NobodyXu:
> I am certainly looking at io-uring based alternatives myself but I agree that's a much larger change and may not fit in this crate.

IIRC there are desires to extract and create a sans-IO core tar crate

PR #396, opened 2025-01-29
Update LICENSE-MIT

maximevtush:
Updated the copyright year from 2014 to 2025 in the LICENSE file

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright (c) 2025 Alex Crichton
+Copyright (c) The tar-rs Project Contributors
 
 Permission is hereby granted, free of charge, to any
 person obtaining a copy of this software and associated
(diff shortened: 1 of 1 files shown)


cgwalters:
What's the motivation here? As far as I am aware copyright dates are meaningless. Ref https://spdx.org/licenses/MIT.html 

The main Rust project doesn't include a year https://github.com/rust-lang/rust/blob/master/LICENSE-MIT 

So my recommendation here is to change this to "Copyright (c) The tar-rs Project Contributors" and we'll never have to change it again.

cgwalters (changes requested):
Marking reviewed

cgwalters (commented):
Just a sanity check @alexcrichton are you OK with this?

alexcrichton:
👍

PR #397, opened 2025-02-11
Make it compile with wasm32-unknown-emscripten

ndmitchell:
Make it so that emscripten works. Now you can do:

```
rustup target add wasm32-unknown-emscripten
cargo build --target=wasm32-unknown-emscripten
```

And it works. Unlike the other WASM target, emscripten configures itself to work with both `unix` and `wasm`, so we have to avoid the unix branch.

A more recent variant of the patch in #301. For that PR, it was suggested that cfg_if would be a reasonable alternative. I'm happy to go that route if you want, but found that it makes things more complicated (to my eye). But maintainers choice.

### DIFF
--- a/src/entry.rs
+++ b/src/entry.rs
@@ -603,7 +603,7 @@ impl<'a> EntryFields<'a> {
                 ::std::os::windows::fs::symlink_file(src, dst)
             }
 
-            #[cfg(unix)]
+            #[cfg(all(unix, not(target_arch = "wasm32")))]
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Would you be up to adding an entry in the CI matrix to compile for emscripten? I'm not sure how to do that myself. If you're not sure though that seems reasonable to leave out too.

ndmitchell:
I'm working on that on a separate branch and will follow up with that later (it's not super easy, but have something hacky working)

alexcrichton:
Ok!

PR #398, opened 2025-02-13
Add WASM CI

ndmitchell:
This fixes the warnings that appear with WASM, then adds CI for it. I initially tried putting it in the matrix, but it is just too different (it's a target, not a toolchain) so doing it as an entirely separate test worked better.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,6 +35,14 @@ jobs:
       run: sudo -E $(which cargo) test
       if: ${{ matrix.os == 'ubuntu-latest' }}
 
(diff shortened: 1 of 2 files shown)


ndmitchell:
Follow up from https://github.com/alexcrichton/tar-rs/pull/397. Would be great if we could get a release after this is merged, so that the version on crates.io supports WASM.

alexcrichton:
Thanks!

alexcrichton:
It's been, uh, a week. I got to https://github.com/alexcrichton/tar-rs/pull/399 though!

PR #414, opened 2025-09-27
Fix lints

cgwalters:


xzfc:
The current MSRV is 1.60/1.63[^1]. Usage of `Error::other()` raises it to 1.74, and `std::iter::repeat_n` raises it to 1.82. Do you have any particular opinion on which MSRV should we stick to?

Some datapoints and discussions:
- [burntsushi](https://old.reddit.com/r/rust/comments/1jmcv5v/a_rant_about_msrv/mkcj760/) suggests N-9 (~1 year) for ecosystem crates, where `N` is the latest stable Rust release.
- the `time` crate uses N-2 policy for public changes and N-4 for internal changes. https://www.github.com/time-rs/time/discussions/535
- https://lib.rs/stats#rustc has MSRV statistics for crates in the registry.
- https://www.github.com/rust-lang/libs-team/issues/72
- https://www.github.com/rust-lang/api-guidelines/discussions/231

[^1]: checked using
      ```
      RUSTC_BOOTSTRAP=1 cargo +1.60 generate-lockfile -Z minimal-versions
      cargo +1.60 check --no-default-features
      cargo +1.63 check
      ```

xzfc (approved):

PR #420, opened 2025-11-11
fix: expose `DETERMINISTIC_TIMESTAMP`

weihanglo:
This exposes `DETERMINISTIC_TIMESTAMP` so for people creating tar entries on the file (i.e., not having file on disk so have no access to `fs::Metadata` to call [`Header::set_metadata`](https://docs.rs/tar/0.4.44/tar/struct.Header.html#method.set_metadata).

This is wanted by <https://github.com/rust-lang/cargo/issues/16237>, though in Cargo we can probably hardcode that timestamp.


weihanglo:
Thanks!

weihanglo:
While I don't think there is any urgency, do we want to publish a minor version for this?

Issue #422, opened 2025-11-18
Initialize `Header::entry_type` to `EntryType::Regular`


e-nomem:
When a header is initialized it currently sets all the underlying bytes to null. This happens to match the entry type `AREGTYPE` which is the 'legacy' regular file type. Any new/modern tar file should use entry type `REGTYPE` (`b'0'`) instead though, which is supposed to be otherwise identical in every way.

This difference came up in `maturin` where some code paths were not explicitly setting the entry type and it uncovered a bug in the python standard library `tarfile` module that is triggered only for `AREGTYPE` entries. See https://github.com/PyO3/maturin/issues/2855 and https://github.com/python/cpython/issues/141707.

It seems to me that end users are most likely creating entries with the legacy type because of a failure to set the type explicitly rather than a conscious choice to use it.

Proposal: Initialize the headers to `REGTYPE` by default and expose `AREGTYPE` in the `EntryType` enum so that users that actually want that mode can explicitly set it.

PR #423, opened 2025-11-21
Initialize headers by default to EntryType::Regular

e-nomem:
Implementation for https://github.com/alexcrichton/tar-rs/issues/422

This is a breaking change to the library.

### DIFF
--- a/src/entry_type.rs
+++ b/src/entry_type.rs
@@ -5,6 +5,8 @@
 /// distinguish between types of content.
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub enum EntryType {
+    /// Legacy Regular file
+    LegacyRegular,
     /// Regular file
     Regular,
     /// Hard link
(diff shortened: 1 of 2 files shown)


xzfc:
There are two different changes:
1. Split `EntryType::Regular` into `LegacyRegular` and `Regular`.
2. Use `b'0'` instead of `b'\0'` by default in `new_gnu` and `new_ustar`.

In the wild (see [grep.app](https://grep.app/search?f.lang=Rust&case=true&words=true&q=EntryType%3A%3ARegular)) I see many uses of checks like `if header.entry_type() != tar::EntryType::Regular`. The first change would silently break such code, i.e. no compile error, but runtime behavior would change. Can it be designed in a non-breaking way?

The second change is less drastic, but still I'd prefer not to change the output across `tar-rs` versions within `0.4.x` version range.

e-nomem:
I suppose we could make `EntryType::Regular` into a tuple or struct variant with some information about if it's a 'legacy' regular file or not? That would break any code matching against `EntryType::Regular` so it doesn't silently fail.

xzfc:
That still would be a a semver violation.

e-nomem:
Yeah, this PR basically _has_ to be breaking because of the change in the encoded default entry type.

PR #424, opened 2025-11-21
Pass an optional `mtime` param to `HeaderMode::Deterministic`

e-nomem:
This `mtime` value will be used instead of `DETERMINISTIC_TIMESTAMP` when provided. This is useful e.g. to pass in `$SOURCE_DATE_EPOCH` to support reproducible builds.

This is a breaking change. Given that `HeaderMode` is marked as `non_exhaustive`, it is possible to make this a non-breaking change by adding a new mode variant that uses the mtime override instead.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -4,6 +4,7 @@ use std::os::unix::prelude::*;
 use std::os::windows::prelude::*;
 
 use std::borrow::Cow;
+use std::cmp;
 use std::fmt;
 use std::fs;
 use std::io;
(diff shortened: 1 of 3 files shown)


e-nomem:
I added a new `HeaderMode::Override` variant that has non-exhaustive struct with individual field overrides. AFAICT the design addresses the use cases from all the linked issues.

It feels a touch verbose but one of the upsides is that both `HeaderMode::Complete` and `HeaderMode::Deterministic` can be expressed as overrides so it does unify the `HeaderMode` handling logic a bit.

Issue #413, opened 2025-09-20
Tar Supports Absolute Paths


anonhostpi:
This snippet should be configurable:

https://github.com/alexcrichton/tar-rs/blob/a1c3036af48fa02437909112239f0632e4cfcfae/src/header.rs#L1565-L1581

Tar supports absolute paths and is used by many build, install, and packaging systems to ensure patches to dependencies living at _known file paths_ are applied when extracting. It's also often used for snapshotting portions of your file system as well.

GNU tar is quite explicit about this, but if I'm correct I believe all tar variants should have support for absolutes: https://www.gnu.org/software/tar/manual/html_node/absolute.html

GNU tar added the `--absolute-names` to make absolute path support explicit so that unintentional directory traversal could be blocked.

zxvfc:
Addressed in PR: #426

anonhostpi:
> Addressed in PR: [#426](https://github.com/alexcrichton/tar-rs/pull/426)

Solid PR kind sir. Added some feedback

zxvfc:
Seems can be closed due to merged: #426

PR #426, opened 2025-12-25
Add support of absolute paths

### LINKED ISSUE #326: Is the absolute path invalid?
hello,  I wanna ask a question,
`tar_file_builder.append_path(&tmp_file_path).unwrap();`
the tmp_file_path can be set absolute path?

zxvfc:
Add `preserve_absolute` to builder's options to let it add files with absolute path.

By default `preserve_absolute` sets false, to keep current behavior, but it can be set as true via passing `true` to builder's method `preserve_absolute`.

Closes: #326, #413

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -23,6 +23,7 @@ pub struct Builder<W: Write> {
 #[derive(Clone, Copy)]
 struct BuilderOptions {
     mode: HeaderMode,
+    preserve_absolute: bool,
     follow: bool,
     sparse: bool,
 }
@@ -35,6 +36,7 @@ impl<W: Write> Builder<W> {
         Builder {
             options: BuilderOptions {
                 mode: HeaderMode::Complete,
+                preserve_absolute: false,
                 follow: true,
                 sparse: true,
             },
@@ -48,6 +50,11 @@ impl<W: Write> Builder<W> {
     /// does _not_ apply to `append(Header)`.
     pub fn mode(&mut self, mode: HeaderMode) {
         self.options.mode = mode;
+    }
+
+    /// Peserve absolute path while creating an archive
+    pub fn preserve_absolute(&mut self, preserve: bool) {
+        self.options.preserve_absolute = preserve;
     }
 
     /// Follow symlinks, archiving the contents of the file they point to rather
@@ -179,7 +186,8 @@ impl<W: Write> Builder<W> {
         path: P,
         data: R,
     ) -> io::Result<()> {
-        prepare_header_path(self.get_mut(), header, path.as_ref())?;
+        let allow_absolute = self.options.preserve_absolute;
+        prepare_header_path(self.get_mut(), header, path.as_ref(), allow_absolute)?;
         header.set_cksum();
         self.append(header, data)
     }
@@ -222,7 +230,8 @@ impl<W: Write> Builder<W> {
     where
         W: Seek,
     {
-        EntryWriter::start(self.get_mut(), header, path.as_ref())
+        let allow_absolute = self.options.preserve_absolute;
+        EntryWriter::start(self.get_mut(), header, path.as_ref(), allow_absolute)
     }
 
     /// Adds a new link (symbolic or hard) entry to this archive with the specified path and target.
@@ -267,7 +276,8 @@ impl<W: Write> Builder<W> {
     }
 
     fn _append_link(&mut self, header: &mut Header, path: &Path, target: &Path) -> io::Result<()> {
-        prepare_header_path(self.get_mut(), header, path)?;
+        let allow_abolute = self.options.preserve_absolute;
+        prepare_header_path(self.get_mut(), header, path, allow_abolute)?;
         prepare_header_link(self.get_mut(), header, target)?;
         header.set_cksum();
         self.append(header, std::io::empty())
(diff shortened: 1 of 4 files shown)


anonhostpi (commented):
Solid simple fix!

For testing, do you think it would be worth it to generate a tar with an absolute path in it using an external tar packer (probably GNU tar, since they are explicit with their absolute path support), unpack it with this crate, then verify the file actually exists?

Would require spawn/cli calls from the test process, but would be comprehensive.

Pathal:
Another solution would be to leverage crate features.

Adding to the `header.rs::copy_path_into_inner` function's match, you don't have to cascade the flag through all that code and make people explicitly opt into "less safe" tar-balls at the dependency level.

```rust
#[cfg(not(feature = "absolute-paths"))]
(Component::Prefix(..), false) | (Component::RootDir, false) => return Err(other("paths in archives must be relative"));
```

and

```rust
#[cfg(feature = "absolute-paths")]
(_, false) => {}
```

I'm also working on a k8s project where leveraging tar-rs with absolute paths would come in handy to (ironically enough) improve pipeline security.

zxvfc:
Hi @anonhostpi! What do you think? Actually @Pathal suggested pretty clean solution (probably even cleaner then mine, since there is no need to pass around `allow_absolute` flag). Does it make sense to refactor my implementation to that?

anonhostpi (approved):
> Hi @anonhostpi! What do you think?

Appreciate the added test. It will ensure that the library is at the same competency as other projects.

> Actually @Pathal suggested pretty clean solution (probably even cleaner then mine, since there is no need to pass around allow_absolute flag). Does it make sense to refactor my implementation to that?

Yes and no. Using feature flags offloads tech debt and forces *the developer* to decide on opt-in, not the end-user.

I think @Pathal's solution is a good one, but it needs to be documented, that if someone uses `tar-rs` to develop a _rust-based tar CLI (or other tar client) competitive with GNU tar_:
- _not only_ do they have to enable the feature-flag, but...
- *should* implement a feature-gate (much like your initial implementation) in _their_ code to prevent insecure packing and unpacking after enabling the feature-flag

In short, I think it is good, but we should be clear on why we decided that, and what debt is offloaded to the end-developer.

cgwalters (changes requested):
I said this elsewhere but what I think we want at some point is to split out a "tar-core" crate that's "sans io" and allows constructing nearly arbitrary tar streams (but of course still giving high level sugar by default to handle e.g. GNU long links and such).

Anyways I think the biggest blocker for this PR as is is that it's a semver break in changing the signature of at least` pub fn append_data` and I think that would require a lot of nontrivial ecosystem churn for a relatively niche use case.

I think we just need the global option, there's no need to allow it per-data addition?

zxvfc:
Thanks @cgwalters for your comments. Yes, you are right, it is really breaking semantic, which is not very good.

I'll see how to rework it and come back with a new changes. :)

zxvfc:
Hello @anonhostpi, @cgwalters!

As promised, I'm back with some rework.
I've removed the `allow_absolute` parameter from Builder's public methods (`append_data` and `append_writer`). These methods now use the internal `self.options.preserve_absolute` setting instead, which avoids the semver break you mentioned.

However the `Header::set_path()` and `Header::set_link_name()` methods still have the `allow_absolute` parameter added. I kept this because: Header seems pretty low level; The Header structure doesn't have an options and I'm not sure that it's good to change it.

Anyway, let me know what do you think about it. 
Thanks!

cgwalters:
https://github.com/alexcrichton/tar-rs/pull/429 adds a CI check for this I'd like to be gating.

> However the Header::set_path() and Header::set_link_name() methods still have the allow_absolute parameter added.

Anyways, to avoid the semver break - and I know it'd be a bit painful but it seems tractable, we could add `Header::set_raw_path()` that literally accepts any bytes.

I did something much like this btw in https://github.com/alexcrichton/tar-rs/pull/274

anonhostpi (approved):
Changes since _last review_ are sylistic, so my review remains the same. LGTM 👍

zxvfc:
@cgwalters, got it! Will post an update soon!

cgwalters:
> Otherwise, Header::set_raw_path will set the path avoiding all checks, which is a bit too much for the case when needed just to set an absolute path.

What other checks are there? I guess I don't have a *really* strong opinion here...ultimately anyone who wants to write arbitrary tar headers can now also use  https://docs.rs/tar-core/latest/tar_core/builder/index.html which very intentionally supports writing whatever bytes one wants as the path.

zxvfc:
@cgwalters 

> What other checks are there?

As far as I can see, `copy_path_into_inner` still enforces: no `..` (parent dir) components, no `null` bytes, non-empty path, and valid encoding — all regardless of `allow_absolute`. The only check that `set_path_absolute` skips is the rejection of `RootDir` `/ - Prefix` components.

My point is that if a programmer wants to set an absolute path, they shouldn't need to bypass *all* validation to do it — `set_path_absolute` lets them opt into absolute paths explicitly while keeping the other safety checks intact. A truly "raw" setter that skips everything (including the `..` traversal check) feels like a separate, lower-level concern, which `tar-core` already covers anyway.

cgwalters (approved):
Ok, fair enough!

PR #449, opened 2026-04-19
Add `Archive::unpack_mapped`

RemiBardon:
(Follow-up to #448, merge first.)

This allows mapping paths while extracting, if #448 gets merged. See this other PR for rationale.

We can discuss implementation, as I’m aware I introduced some genericity which might affect performance / binary size / compile time. There are no benchmarks so I couldn’t know if my changes had an impact. It _should_ not since most people will never use `unpack_mapped` and `_unpack` will get monomorphized to the exact same function as before, with the compiler likely optimizing away the newly added line as it’s a no-op. And in the rare case where one needs `unpack_mapped`, they’d likely have a single call site so `unpack_mapped` wouldn’t bloat the binary. But that’s just my opinion, I’d get it if you want to use dynamic dispatch instead.

Oh and also feel free to change the method name, I had no inspiration 🙆‍♂️

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -120,7 +120,18 @@ impl<R: Read> Archive<R> {
     /// ```
     pub fn unpack<P: AsRef<Path>>(&mut self, dst: P) -> io::Result<()> {
         let me: &mut Archive<dyn Read> = self;
-        me._unpack(dst.as_ref())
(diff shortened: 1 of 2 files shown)


RemiBardon:
(FYI I ended up not needing this as I need to iterate over entries myself. You can close this PR if you feel like it’d be too intrusive.)

